Issue #11, opened 2014-08-30
Add meta information like values and labels on SVG elements in DOM


gionkunz:
We should store the information in the DOM too. I guess some of this goes hand in hand with the accessibility work.

Like this it will even become easier for end users to create their own behavior on charts.

phun:
Should we be using `desc` elements instead since the data attribute isn't defined for SVG in W3C recommendations? More info here: http://stackoverflow.com/questions/15532371/do-svg-docs-support-custom-data-attributes

gionkunz:
Jup, looks like it. Wasnt even thinking about the possibility that it doesnt allow data attributes as i thought its a regular dom node. Lets also wait for the BBC guys from issue https://github.com/gionkunz/chartist-js/issues/2 if they come up with something into this direction. Thanks for pointing out!

JamieKnight:
Hiya Chaps,

Just picking this up now. Its a new area for me so I am very open to suggestions. 

I have been thinking, it might be good to start by reviewing the chart types and for each chart type discussing what the best user experience would be for a non sighted user. From there we can then figure out if it would be best to make the SVG more accessible or if we can refine the markup used underneath to provide an accessible alternative. 

For example (and based on little more that optimistic poking, so shout if this sounds crazy):

### Pie chart

IMHO the best UX for non sighted users would be some labels each with the percentage. This could be achieved either a list which is transformed (yay progressive enhancement) or via ARIA labels in the SVG. We could easily support either depending on if the library pulls data from the DOM, or renders data passed via JS.

### Line Chart

Ideally, a line chart would read it's values in a meaningful way. For example imagine a chart of items each with a price... if it could be read as "Item fishtank, Cost £150" that would be ideal.

The table elements, when used to thead and its friends support this type of reading. So for the line chart it may be best to leave the SVG as invisible, and instead focus on the alternative markup where the best user experience can be achieved.

How does this approach sound? If we could do a pass through each chart type and agree the desired UX, we can then incrementally improve and tackle the low hanging fruit first.

I'm very much thinking aloud, so shout if this sounds crazy

Cheers,

Jamie + Lion

JamieKnight:
Hello Again,

I have just had a reread of the docs (http://gionkunz.github.io/chartist-js/api-documentation.html). It seems we only have the three chart types:

1: Pie Chart
2: Line Chart
3: Bar Chart

[...]

PR #28, opened 2014-09-04
added new attributes for the line series

ghost:
because we need new attributes for a "tooltip", we have added the attributes "x" and "y" to the line element


gionkunz:
Hi there, first thanks for your contribution. I really appreciate that you put your changes back to the community.

Please give the contribution.md file a shot and check the guidelines. In pull requests, you should never include the libdist.

Also I believe we need to solve this differently as this does not comply with the standard see https://developer.mozilla.org/en-US/docs/Web/SVG/Element/line for valid attributes

We can track the effort under #11 

I'll close the request. Please reopen a new one if you'd like to contribute to the efforts under #11.

Cheers
Gion

Issue #4, opened 2014-08-29
Hover states


herkyl:
It would be nice to have hover data on PC and touch tooltips on mobile.

gladson:
:+1:

gionkunz:
Thanks for opening an issue. Actually I was thinking about adding an API for such things a lot but I came to the conclusion that we won't add functionality like this to Chartist.

The reason for not including this is simply that, while using inline-svg, a proper, semantic and annotated DOM, it's so simple to add this functionality yourself. It would pollute the library with stuff that is not the main concern of Chartist, which is only the drawing of the chart.

If you like to have behavior added to your charts you can do that with a few simple lines with using the DOM library of your choice.

Check out this JSBin (which is actually a very bad example because you should to this kind of visual only behavior with CSS, but it shows how easy it is)

http://jsbin.com/yisaye/1/

herkyl:
Thanks for the reply. You should put some examples of how to add features like this to the docs. :)

gionkunz:
Sure I will! :-) And thanks for your feedback! This will be in the advanced usage example of chartist.js. Generally the documentation is something I (who not?!?) need to catch up.

Btw. your issue made me notice that there is a possible bug and the dots are hidden behind the lines which should be the other way around I guess ;-)

gionkunz:
I wanted to note on this issue that there is now full support for functionality like this with the new custom namespace attributes. Check the advanced example here http://gionkunz.github.io/chartist-js/getting-started.html#behavior-with-jquery

Cheers
Gion

arechsteiner:
@gionkunz: Would you reconsider adding hover and tool tips to Chartist out of the box? I think it's an essential feature for charts displayed in the browser. I really like the approach you took using inline SVG and CSS and you've done a great job with the documentation (the live-edit code examples are something I will most definitely steal ;)

The only fly in the ointment is that tool tips on hover are not included. I think this might as well be the most popular chart library on GitHub otherwise.

Anyway just my two Rappen. Cool project either way.

gionkunz:
Did you check the chartist tooltip plugin?

arechsteiner:
Yes, I followed the link to the docs example but it seems broken:

![brokenchartisttooltip](https://cloud.githubusercontent.com/assets/1199486/15799647/95322aac-2a63-11e6-8cad-fb3e66792046.png)

[...]

PR #30, opened 2014-09-09
added simple tooltips to charts

dominicrico:
Hi,

I added simple tooltips on hover over the points, bars or slices with the current value, pre- and/or suffix text.


willsoto:
I think this was already addressed in #4 with the conclusion that this sort of functionality will not be included in the core, as it is trivial to add yourself without the library.

Also, `libdist` should not be included in the PR. It is stated in [CONTRIBUTING.md](https://github.com/gionkunz/chartist-js/blob/develop/CONTRIBUTING.md) albeit in a rather inconspicuous place.

dominicrico:
Sorry, haven't read that issue until now and the also sorry for the libdist
in PR.

willsoto:
No big deal. It should be noted I am not a core contributor or anything, perhaps the owner will have a different thought about it.

Thanks for the PR though, glad to see this repo is getting attention.

dominicrico:
It's great to have a easy chart library like this and in my opinion many
people would love to have "build in tooltips". This was my first try to
contribute and I learned a lot :D

luanmuniz:
Hi guys, im new here, but here is my opinion:

You should do this feature as a library that extend chartist.js
A good way to explain my opinion is to say: This feature extend chartist.js as jquery-ui (or any plugin) extend jquery.

You can make your own repository with this code and make this feature as a extend for chartist.js
something like chartist-tooltip.js
That way, its not inside the core, but we can use in a easy way.
and OBVIOUSLY mention it here! o/

What do you guys think?

dominicrico:
Yes, that sounds great and I was thinking about that too.

I will do this anyway, because it's a good training for me.

gionkunz:
I agree to most points of this discussion :+1: 

Please check the advanced example I've just pushed into the gh-pages that contains a first topic of how to add behavior to Chartist from outside.

http://gionkunz.github.io/chartist-js/getting-started.html#adding-behavior-to-your-charts

Cheers
Gion

vlasvlasvlas:
i just test the online demo found here http://gionkunz.github.io/chartist-js/getting-started.html#adding-behavior-to-your-charts the last one "ADDING BEHAVIOR TO YOUR CHARTS" with tooltips,

and at localhost i got this error:

```
Uncaught ReferenceError: $chart is not defined 

line 33: 
var $toolTip = $chart
```

my code:

```
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="libdist/chartist.min.css">
    <script src="libdist/chartist.min.js"></script>
  </head>



  <body>

<div class="ct-chart ct-perfect-fourth"></div>
  <div class="tooltip"></div>
  </body>


  <script>
  Chartist.Line('.ct-chart', {
  labels: ['1', '2', '3', '4', '5', '6'],
  series: [
    {
      name: 'Fibonacci sequence',
      data: [1, 2, 3, 5, 8, 13]
    },
    {
      name: 'Golden section',
      data: [1, 1.618, 2.618, 4.236, 6.854, 11.09]
    }
  ]
});

// $chart is set to a jQuery wrapper around the chart container div
var $toolTip = $chart
  .append('<div class="tooltip"></div>')
  .find('.tooltip')
  .hide();

$chart.on('mouseenter', '.ct-point', function() {
  var $point = $(this),
    value = $point.attr('ct:value'),
    seriesName = $point.parent().attr('ct:series-name');

  $point.animate({'stroke-width': '50px'}, 300, 'easeOutQuad');
  $toolTip.html(seriesName + '<br>' + value).show();
});

$chart.on('mouseleave', '.ct-point', function() {
  var $point = $(this);

  $point.animate({'stroke-width': '20px'}, 300, 'easeOutQuad');
  $toolTip.hide();
});

$chart.on('mousemove', function(event) {
  $toolTip.css({
    left: event.offsetX - $toolTip.width() / 2 - 10,
    top: event.offsetY - $toolTip.height() - 40
  });
});
</script>

</html>
```

whats my problem?

dominicrico:
You have to declare $('.ct-chart') = $chart, because your $chart refers to
nothing.

vlasvlasvlas:
Ok.. but where?

What i mean with these is, here is a public web example page http://gionkunz.github.io/chartist-js/getting-started.html#adding-behavior-to-your-charts and a simple copying of the code doesnt do the job. Maybe for better usability of examples the "show code" preview must be more real world copying code and testing. (working example)

thanks! :)

dominicrico:
Just do it before the var $tooltip = $chart ... It has to look like
var $chart = $('.ct-chart');
var $tooltip = $chart ...

vlasvlasvlas:
Thanks dominicrico! i do mean that working example of public "show the code" on the website would be nicer if working examples works just by copying them to localhost :) thanks for everything!! i'm so happy with this charting library!

dominicrico:
No problem, you are welcome! Great to here it works now for you.

PR #33, opened 2014-09-09
Matching responsive scale between docs and code

CalebKester:
I noticed that about half of the documented responsive scales weren't actually in the code (and the code had ct-square which wasn't in the documentation).

This is only fixed in the .scss and the yml. Not in the .css


gionkunz:
:+1:  Great! Nicely spotted and thanks for the contribution! This will go straight in.. :shipit:

PR #50, opened 2014-09-21
Improved test coverage

valotas:
Added instanbul as a dependency and some tests for `chartist.core.js`. Fixed already some edge cases found during testing


ilanbiala:
@valotas @gionkunz is this PR still necessary?

gionkunz:
I appreciated the efforts of @valotas but this PR is so outdated..... I'll close the PR and maybe steal some code if we decide to go for full blown test runner.

PR #57, opened 2014-09-29
Update README.md

jonsadka:
Removed duplicate word 'is' on line 26.
Line 26: "problem which is is already solved" --> "problem which is already solved"


gionkunz:
Great! Thanks @jonsadka ! :-)

PR #61, opened 2014-10-06
Fix for donut chart when a series only has one non-zero value

djlimbs:
If a data series had only one non-zero value that was not the first element, such as [0, 0, 23, 0, 0], it would incorrectly set the startAngle. This fix checks the series first then updates the startAngle calculation to take that into consideration.

### DIFF
--- a/source/scripts/chartist.pie.js
+++ b/source/scripts/chartist.pie.js
@@ -166,6 +166,9 @@
         y: chartRect.y2 + chartRect.height() / 2
       };
 
+      // Check if there is only one non-zero value in the series array.
+      var hasSingleValInSeries = data.series.filter(function(val) { return val !== 0; }).length === 1;
+
       // Draw the series
       // initialize series groups
       for (var i = 0; i < data.series.length; i++) {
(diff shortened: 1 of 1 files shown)


source/scripts/chartist.pie.js
```diff
@@ -166,6 +166,9 @@
         y: chartRect.y2 + chartRect.height() / 2
       };
 
+      // Check if there is only one non-zero value in the series array.
+      var hasSingleValInSeries = data.series.filter(function(val) { return val !== 0; }).length === 1;
```
gionkunz: Cool stuff :-) Thanks for working on this and using Array extras! In Chartist we follow the convention to write function bodies on a new line. Can you put the return statement on a new line something like this?

``` javascript
var hasSingleValInSeries = data.series.filter(function(val) {
  return val !== 0; 
}).length === 1;
```

djlimbs: Yes, no problem, I'll update it and re-submit. Thanks!

gionkunz:
Just tested this and it looks good :+1:

djlimbs:
Awesome!

Issue #70, opened 2014-10-16
Return values for chart constructors


willsoto:
Would it be possible to return some sort of chart object that contains information related to the newly constructed chart?

For example, return the final values for the config objects that were passed in once they are merged with the defaults.

A little context for why I feel this would be good for 3rd party libs (and to make sure I'm not falling victim to the [XY problem](http://meta.stackexchange.com/a/66378)):

I am trying to find a way to handle events (e.g., click, mouseover, etc) in an angular way. The best way to do this, I think, would be as you suggested in #36 ([comment](https://github.com/gionkunz/chartist-js/issues/36#issuecomment-55454870)), which is to use nested directives. If I were to go this route, it would require a lot more effort, but it would be the **right** way to go. I may explore this in a branch and get your feedback...perhaps this can be a dual endeavor. I would want to do this in such a way though that the entire library is not being re-written.

The other option is get the relevant elements and attach emitters (maybe, still working through this). But this presents it's own problems in that the class names can be changed by the user and I don't want to have logic in the directive to figure out whether or not this happened. This means that the directive would have to maintain knowledge about the default values, and if they change then I'd have to update the wrapper, etc. I'd rather just read off the final value from the config object. 

Do you have any thoughts on this? I'm open to suggestions on the best way to get this done.

Thanks!

gionkunz:
Hey Will

I get the idea and thanks for explaining with so much detail. There are just custom events in Chartist where they are actually needed, or in other words, where you don't have events already by the standard (matchMedia, DOM events etc). This means for click, hover etc. we should also leverage the DOM events in ng-chartist.js. For all events that can only come from Chartist we have the EventEmitter inside of chartist to which you can bind using the `on` function. For the current options there is the `optionsChanged` event you can use. Its emitted here: https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.core.js#L596-L599 This event is also re-emitted if the options change due to the media query listeners and the override mechanism in Chartist. The problem currently is that Chartist does the first draw and initialization when you create the Chart with the "constructor" function. By then it's already too late to listen for any events. For this purpose I was moving the initial `createChart` call into a setTimeout so it performs sometime after the current call stack finished which gives you the possibility to register for draw events right after you created the chart and they will be triggered. The "problem" currently is that the `optionsProvider` is initialized still in the call stack where you initialize the chart. https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.line.js#L384-L389

[...]

PR #71, opened 2014-10-17
Move optionsProvider into setTimeout

willsoto:
This moves the `optionsProvider` call into the `setTimeout` as discussed in #70


willsoto:
Haha yeah...lol

PR #76, opened 2014-10-24
Corrected typo in barchart options

SteveEdson:


gionkunz:
Hi there! Thanks for the pull request. gh-pages is generated from the source so you should make a pull request there. https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.bar.js#L237

Issue #80, opened 2014-11-10
Custom event for 'drawEnd'


jordanwade:
Looking through the source and the examples, I'm trying to establish how to create a custom event handler via `addEventHandler` when the `draw` of the chart has finished. I can see how to create the event handler and `emit` it, but how exactly would your trigger the `emit` within the source creation of the `svg`?

gionkunz:
Hi there. Can you give an example of what you'd like to achieve? This would make it easier to imagine what you're looking for. Do you want an event when the whole drawing is done? Or after each element is drawn?

jordanwade:
I would like an event when the whole drawing is complete. I am trying to append  a `defs` element inside the drawn `svg` in order to create a `lineargradient`, but the `draw` event will trigger the append when each element is drawn. Essentially, I'm hacking it together with this:

``` javascript
var appended = false;

chart.on('draw', function(data) {
  if (!appended) {
    // Do something once...
    appended = !appended;
  }
});
```

Obviously, it would be much nicer/cleaner to have an event that is triggered once the drawing is done in order to append the element inside the `svg`.

gionkunz:
what about an event `created` right after the chart was created? The chart updates currently cause a re-creation anyway. In the future we'll probably be able to recycle some of the DOM elements on certain updates but for now a re-creation is fine. In the line chart this would go here: https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.line.js#L175

What do you think?

Cheers
Gion

jordanwade:
Yeah, a `created` event there would be perfect. I just need to manipulate the `svg` _after_ the entire chart is drawn. If you know of a better way to alter the `svg` that is created internally, that would work for my current situation as well.

Thanks so much for the quick reply.

gionkunz:
If you like you could create a pull request for this. I think a `created` event wouldn't hurt :smile: . I can see this can be helpful. For creating gradients, patterns or filters too. We would need to add it to all three chart modules. Here are examples of the other events: https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.line.js#L114-L122 or https://github.com/gionkunz/chartist-js/blob/develop/source/scripts/chartist.core.js#L538-L548

drewbarontini:
Hey, @gionkunz - just want to say I work with @jordanwade and have been working on this issue with him. We're happy to help out and take a stab at creating the event in a Pull Request. I'll work on getting that for you to review. Thanks again for the help!

[...]

PR #81, opened 2014-11-10
"Created" Event

drewbarontini:
## What is this?

This adds a new event called `created`, which is triggered once a chart is finished drawing. This is created for all three charts, and it returns:
- The `bounds`
- The `chartRect`
- The `options`
- The `svg`
## Example usage

``` javascript
  labels: ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday'],
  series: [
    [12, 9, 7, 8, 5],
    [2, 1, 3.5, 7, 3],
    [1, 3, 4, 5, 6]
  ]
});

chart.on('created', function(data) {
  // ...
});
```
## Related issue

[Refs #80]

### DIFF
--- a/source/scripts/chartist.bar.js
+++ b/source/scripts/chartist.bar.js
@@ -120,6 +120,12 @@
         });
       }
     }
+
+    this.eventEmitter.emit('created', {
+      type: 'Chartist.Bar',
+      labels: labels,
+      svg: this.svg
+    });
   }
 
   /**
--- a/source/scripts/chartist.bar.js
+++ b/source/scripts/chartist.bar.js
@@ -122,9 +122,10 @@
     }
 
     this.eventEmitter.emit('created', {
-      type: 'Chartist.Bar',
-      labels: labels,
-      svg: this.svg
+      bounds: bounds,
+      chartRect: chartRect,
+      svg: this.svg,
+      options: options
     });
   }
 
--- a/source/scripts/chartist.line.js
+++ b/source/scripts/chartist.line.js
@@ -175,9 +175,10 @@
     }
 
     this.eventEmitter.emit('created', {
-      type: 'Chartist.Line',
-      labels: labels,
-      svg: this.svg
+      bounds: bounds,
+      chartRect: chartRect,
+      svg: this.svg,
+      options: options
     });
   }
 
--- a/source/scripts/chartist.line.js
+++ b/source/scripts/chartist.line.js
@@ -173,6 +173,12 @@
         }
       }
     }
+
+    this.eventEmitter.emit('created', {
+      type: 'Chartist.Line',
+      labels: labels,
+      svg: this.svg
+    });
   }
 
   /**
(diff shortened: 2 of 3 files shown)


drewbarontini:
@gionkunz - I might need your helping sorting out what data to pass back from the event. Currently, I just send the `type`, `labels`, and `svg`, but the `Chartist.Pie` doesn't have the same `labels`, so I pass back `seriesGroup`. Let me know what you think!

gionkunz:
@drewbarontini, thanks for the great PR! I think so far we have included data in the event that is crucial to get the full context of the event while it occurs. In the case of the drawing events this is the index of the element, the element itself, type of the drawing ('point', 'line', 'bar' etc.).

I think for the chart created event there is almost no special context that you're not already aware from the outside and therefore I think we should stick to stuff from the closure in the create function that could be of a help in the event. For me this would be `bounds`, `chartRect`, `options` (as these were used for that specific drawing). The Pie chart is a bit special because it has no `bounds` as it operates on simple polar coordinates. Also I think we don't need the `type` field as this is only required when you have the same event with different subjects in the same chart (like draw which is triggered for labels, points, lines, slices etc.).

drewbarontini:
Alright, I updated the returned object to include:
- The `bounds`
- The `chartRect`
- The `options`
- The `svg`

I kept the `svg` in there since there is no other place to gain access to the internally created SVG from outside, which makes it useful within the 'created' event. Let me know what you think.

drewbarontini:
And as an example, we used it to test out creating a gradient overlay on the SVG:

``` javascript
function createGradientOverlay(svg, options) {
  var defs = svg.elem('defs');
  var grad = defs.elem(options.gradient.type, options.gradient);
  for (var stop in options.stops) {
    grad.elem('stop', options.stops[stop]);
  }
}

chart.on('created', function(data) {
  createGradientOverlay(data.svg, {
    gradient: {
      type: 'linearGradient',
      id: 'grad1'
    },
    stops: {
      1: {
        offset: '0%',
        style: 'stop-color:red;stop-opacity:1'
      },
      2: {
        offset: '100%',
        style: 'stop-color:blue;stop-opacity:1'
      }
    }
  });
});
```

gionkunz:
Nice! :smile: 

Thanks for this guys! I'll merge that into develop straight away. I'll make a release sometime later this week.

Cheers
Gion

drewbarontini:
:+1: Glad we could help!

nblackburn:
@gionkunz @drewbarontini Is it possible to gain access to the series data from this event?

gionkunz:
The data is exposed directly on the chart object. You can access it with `chart.data`. If you want to modify the data before Chartist is using it to draw, you can use the `data` event that happens before the `create` event.

nblackburn:
My idea was to use this data to find elements over a certain value and apply a certain class to them as part of a plugin i am working on.

gionkunz:
For this you have the ct:value attribute on each bar element. This might not be the right place to discuss further. Plase open a new issue if you need further assistance with your plugin. Cheers gion

PR #83, opened 2014-11-17
Added less files

ghost:
Added less stylesheets.


gionkunz:
Hi there. I think this is a great idea! I personally prefer Sass but having Less files also available in Chartist.js is a great idea. However, I think there should be a way of auto generating the less stylesheets from the Sass / SCSS files. I don't want to find chartist contributors to maintain two different CSS preprocessor sources (considering the relatively small complexity of the SCSS files).

Did you check if https://github.com/jonschlinkert/grunt-refactor could do the job? Otherwise since LESS is covering almost every feature of Sass since 1.6 / 1.7 I guess creating a simple grunt task that uses copy and replace with regex could also do the job.

autarc:
Automation would definately be more flexible - as styles changes will happen and require otherwise additional maintenance. I am planning a similar approach for stylus, which will use [this converter](https://github.com/mojotech/sass2stylus/) to keep the actual sources files in sync. Supporting different pre-processors out of the box makes it easier for people to modify and integrate the library in their projects.

Issue #39, opened 2014-09-14
Make gruntfile more modular so it's easier to spot different tasks


valotas:
It is all about [SoC](http://en.wikipedia.org/wiki/Separation_of_concerns) after all. Right?

Apart from that it will help a lot the newcomers to the project.

gionkunz:
Please specify what you'd like to split? The site is very well separated in the current project and I like the benefit of having just one repo. Currently source/site, source/scripts/site and source/styles/site are relevant to the site. All other resources are relevant for the library.

I know it would also be nice to split it into a fully separated project. But, again, this has also a few drawbacks. For example the development pace is slower as you need to switch projects all the time and compile the library project to make use in the documentation project. 

Given the clean structure with the sites folder I'd not like to go with a separate project at the moment just to keep it compact and fast paced.

Cheers

valotas:
In that case just split the `Gruntfile.js` to server and library.

gionkunz:
Why? :smile: The gruntfile is the dev part of the library. It's there to provide a development environment with live reload, linting etc. and to create the distribution package as well as the documentation site. How would you like to split the Gruntfile? The "server" tasks share many sub-tasks with the library thing. What would make sense, would be to separate the Gruntfile into much smaller files in order to structure it a bit better.

valotas:
Why? Mostly because I have to go through everything there when I only care about core js stuff :)

It makes me nervous when I write javascript and find `assemble` hanging around :)

gionkunz:
hehe :smiley: Well its also nice to have a visual feedback of your javascript when you code ;-) Specially on such a visual project like this. Also assemble is responsible to create the API documentation from your javascript comments, so it's important to have feedback there too :-) I guess you need to conquer your fear :smile: 

I'll rename this issue to something more generic that I agree with.

Cheers

PR #86, opened 2014-11-21
Restructure build tasks

autarc:
Using seperate configuration helps organizing the process, while lazy loading the modules improves the start.


gionkunz:
Hi there! Thanks so much for this PR! :-) This is overdue for a long time ;-) And actually addresses #39 . Do you think it would also make sense to group certain tasks together and have separate files for our own tasks as well? Like here https://github.com/gionkunz/frontstart/tree/master/tasks

For the changelog thing I'm still not 100% sure about what's the best approach. I'm actually not a huge fan of hiding the sausage making in my commit history. I agree that a commit should never break the system, but slowly reveal more and more functionality that, at some point, could also only partially fulfill a whole features requirements. Therefore I do appreciate small commits that group together bundles of logical units that aim towards a final feature. However, I can see the benefits of larger commits too (bisect, code review, fully working features with one commit etc.). I believe it also highly depends how transparent you'd like to be and more small commits can be great for documentation purpose too.

Let's do it like this. I'll use your branch locally and merge it with my pending things. I'll then start to use the angular changelog convention for my commits and also rebase / squash my pending stuff so it conforms more to feature commits. I'll then develop towards 0.4.2 using this format so I can see the pros / cons better.

Anyways thanks for the PR this is really great!

Cheers
Gion

autarc:
Your're welcome :)

I saw the issue and found the monolithic setup annoying as well. Fortunately [load-grunt-config](https://github.com/firstandthird/load-grunt-config#config-grouping) already provides the tools for grouping, so it can be added straightforward.

For the changelog you are probably right - forcing people to follow a specific convention for their commits could be a problem. The problem was just that I had no clue whats the current situation of the library, as common files like CHANGELOG.md or HISTORY.md are missing. Providing a short description about the latest changes in the specific versions/patches would be great.

gionkunz:
I think for now I'll merge the config change but let's stick to a manual changelog.md for now. I'll include one for the last few versions soon. Can you update the serve task stuff with my previous comment? Then I'll merge everything up to this commit but skip 07da672afe81964d11ba87051c8fa4661b804f4c for the moment.

Would be great to do this before any effort on #87 so I can merge it to develop and we can start a feature branch for #87 .

gionkunz:
Hey there, did you test the site after your refactoring? When I start the site now with grunt live, there is some missing styles on the site. Maybe because of the updated deps? critical or Sass? Let me know if you need help.
screencapture-0-0-0-0-9000

autarc:
Sorry, with the update of grunt-sass a [newer version of node-sass/libsass](https://github.com/sindresorhus/grunt-sass/commit/a7907718effdb441013341e68925a7f950666316) is used which breaks with current build. Reversed the change and like to mention the idea of seperating library and website source code in the future (#88).

gionkunz:
Cool, I've pushed to develop (except 07da672)

Issue #88, opened 2014-11-23
Seperate library code from website


autarc:
As Chartist itself is a library and not a service, its confusing to keep the static files for the website documentation along with the JavaScript sources. Since github pages are already used a clean seperation should be reasonable. Plain examples how to use certain features could still be shown in a 'examples' folder. This would also allows to use common terms for the existing structure (e.g. "libdist" was probably choosen instead of "dist" to provide a hint that its the code and not the website build).

gionkunz:
Hi there. The reason why we still stick to this is mainly because Chartist is a very visual library and it's very nice to develop on the library while using the documentation site as a fast feedback for your development changes. If we rip out the documentation site into its own project we'd need to have a way how to develop on Chartist and have a visual representation / sandbox in order to develop more rapidly. Actually I see the point to separate the library from the site, but I do like the benefit of having a living sandbox during development. Also I like the fact the for contributors they only need to grab one repository and they can hack around. 

What if we stick to this but make the separation even more clear? I.E. Also put images, scripts/site and styles/site into site/images, site/scripts and site/styles ?

autarc:
Thats seems to be the dilema of larger libraries and frameworks in general, often handled with a seperate repository just containing the static website. I think we can avoid this overhead by making the seperation more clear as you mentioned. Extracting the `website` specific code for the documentation and also provide `examples` on how to use the library would definately help. [Chart.js](https://github.com/nnnick/Chart.js) has a similar approach, which makes it easy for others to take a look at the source or specific use cases.

PR #89, opened 2014-11-25
Seperate source files

autarc:
As mentioned in #88 - a clear seperation of the source files for the Chartist library and the documentation site. Like in the previous version are name/path configurations defined in the package.json.


gionkunz:
Cool man! :-) I'll review that ASAP!

gionkunz:
Hey thanks for the changes. I applied some minor changes including:
- renamed folder types to charts
- moves JS files of library to src/scripts
- fixes usemin path and removed redundant concat / minify tasks
- fixed dev task to only perform assemble and sass as the raw source js files will be used and no build is required

I'll work on the changelog for the past releases and put your changes into 0.4.2 together with mine.

PR #93, opened 2014-11-26
Fix small error in setup instructions

danieldiekmeier:


gionkunz:
Thanks! It should actually even be `grunt dev`! We just changes the internals a bit in the latest commits and forgot to update the contrib guidelines. Let me quickly fix it and thanks for the PR!

PR #113, opened 2014-12-12
change the order between area and line

deckool:
this way, you can have a stroke for area and also you can see the line above area. in the current version, when you have both strokes the area it overrides the line. which is not good for styling

### DIFF
--- a/dist/chartist.js
+++ b/dist/chartist.js
@@ -1880,7 +1880,23 @@
               pathElements.push('L' + pathCoordinates[l - 1] + ',' + pathCoordinates[l]);
             }
           }
+          
+          if(options.showLine) {
+            var line = seriesGroups[i].elem('path', {
+              d: pathElements.join('')
+            }, options.classNames.line, true).attr({
+              'values': normalizedData[i]
+            }, Chartist.xmlNs.uri);
 
+            this.eventEmitter.emit('draw', {
+              type: 'line',
+              values: normalizedData[i],
+              index: i,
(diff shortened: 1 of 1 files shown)


deckool:
did my last reply enlightened the pointed functionality? you can keep add event handlers to points as they are not touched by changes. is only a matter for designers most, either way you can not style the line because it is bellow area. If in other cases there are difficulties obtaining other performances, please add them here, we can dig.

gionkunz:
Hey there. I've tested your changes again and you're absolutely right. I somehow always get confused that the DOM order of SVG implies the depth and that elements inserted earlier in the DOM are higher up in the layers :-) With your PR the order will be area > line > points but currently it's line > area > points. Thanks for this and your patience!

Cheers
Gion

gionkunz:
I'll make sure this goes in the next minor release, thanks for your contribution.

deckool:
you are welcomed. I also thank you for the good work with chartist!

Cheers,
Decebal

gionkunz:
Crap we merged this into master :-(

gionkunz:
I just realized when I tried to make the new release

deckool:
Now i realized also, looking in the first merge. Is everything good? ,it seem to be.

gionkunz:
Yeah I had to merge a bit back and forth :-) its not the cleanest history but it seems to be fine now :-) So your change is now in 0.6.0. Cheers

deckool:
I'm glad it's ok. :) Cheers

PR #114, opened 2014-12-12
Improved consistency of comments

drdla:
- fixed typos
- consistent capitalization
- respect maximum line width of 120 characters

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -53,24 +53,24 @@
       bounds,
       normalizedData = Chartist.normalizeDataArray(Chartist.getDataArray(this.data), this.data.labels.length);
 
-    // Create new svg element
+    // Create new SVG element
     this.svg = Chartist.createSvg(this.container, options.width, options.height, options.classNames.chart);
 
-    // initialize bounds
+    // Initialize bounds
     bounds = Chartist.getBounds(this.svg, normalizedData, options, 0);
 
     var chartRect = Chartist.createChartRect(this.svg, options);
     // Start drawing
     var labels = this.svg.elem('g').addClass(options.classNames.labelGroup),
       grid = this.svg.elem('g').addClass(options.classNames.gridGroup),
-    // Projected 0 point
+      // Projected 0 point
       zeroPoint = Chartist.projectPoint(chartRect, bounds, [0], 0);
 
     Chartist.createXAxis(chartRect, this.data, grid, labels, options, this.eventEmitter, this.supportsForeignObject);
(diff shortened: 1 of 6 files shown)


src/scripts/charts/bar.js
```diff
@@ -97,7 +97,8 @@
           bar;
 
         // Offset to center bar between grid lines and using bi-polar offset for multiple series
-        // TODO: Check if we should really be able to add classes to the series. Should be handles with Sass and semantic / specific selectors
+        // TODO: Check if we should really be able to add classes to the series.
+        // Should be handled with Sass and semantic / specific selectors
```
gionkunz, line 101: They there. I think introduce line breaks in TODO comments is a bad idea as some IDEs have great support for TODO / task lists but this breaks the support. Generally we should check that lines don't get too long but the times of consoles with 80 characters width and no mouse to scroll horizontally are over :-)

src/scripts/charts/line.js
```diff
@@ -1,5 +1,7 @@
 /**
- * The Chartist line chart can be used to draw Line or Scatter charts. If used in the browser you can access the global `Chartist` namespace where you find the `Line` function as a main entry point.
+ * The Chartist line chart can be used to draw Line or Scatter charts.
```
gionkunz, line 2: Currently dox does not ignore line breaks and creates a new paragraph for every line break so introducing line breaks in dox comment blocks is not a good idea.

src/scripts/charts/line.js
```diff
@@ -97,13 +99,13 @@
         pathCoordinates = [],
         point;
 
-      for (var j = 0; j < normalizedData[i].length; j++) {
+      for(var j = 0; j < normalizedData[i].length; j++) {
```
gionkunz, line 102: Thats mainly style preference. I prefer the other way. If you'd like to introduce consistency you should check that this is treated all the same everywhere in the code.

src/scripts/core.js
```diff
@@ -425,7 +426,7 @@ var Chartist = {
         height = options.axisX.offset,
         pos = chartRect.x1 + width * index;
 
-      // If interpolated value returns falsey (except 0) we don't draw the grid line
+      // If interpolated value returns falsy (except 0) we don't draw the grid line
```
gionkunz, line 429: Well... that's hard to say :smile: http://www.google.com/trends/explore#q=falsy%2C%20falsey

gionkunz:
Hey there :-) Thanks for your PR but there are too many things I don't agree with. I'll close this PR. Please check the individual comments.

PR #134, opened 2015-01-08
Add options param to update() to extend options

christianseel:
This little addition makes it possible to extend the options with the update() function.

This is useful if you need to change the labelInterpolationFnc together with changed data.
Tested in my current project without issues.


gionkunz:
Hey man thanks for the PR thats great! You should call `removeMediaQueryListeners` on the old options provider before you instanciate a new one.

christianseel:
@gionkunz added with 8b500a2

gionkunz:
Thanks for this Christian!

PR #140, opened 2015-01-12
bower ignore: add "tasks"

coderaiser:
Add Grunt folder `tasks` to `ignore` in `bower.json`.


gionkunz:
Cool thanks! That was missed when we did the structure refactoring!

PR #142, opened 2015-01-13
fix(core) createSvg: add container check before remove

coderaiser:
`svg` could be not direct child of `container`, for example:

``` html
<body>
    <div id="container">
        <svg></svg>
    </div>
</body>
```

``` js
var svg = Chartist.createSvg(document.body)
```

this would fail with:

``` js
 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -165,7 +165,7 @@ var Chartist = {
     height = height || '100%';
 
     svg = container.querySelector('svg');
-    if(svg) {
+    if(svg && container === svg.parentElement) {
       container.removeChild(svg);
     }
 


tylergaw:
I ran into this error today too. In the element I was rendering the chart, I have a `<div>` that contains an `<svg>` element to display a loading graphic. The `createSvg` attempts to remove the loader `<svg>`, but errors since it's not a child of the parent element.

Here's an example:
http://jsfiddle.net/thegaw/woept24e/1/

gionkunz:
This was solved with https://github.com/gionkunz/chartist-js/pull/147

PR #146, opened 2015-01-15
Only remove chartist svg on createSvg

tylergaw:
This is a slightly different take on #142

@gionkunz I tried the approach you mentioned with `container.__chartist__.svg` but couldn't get it to work. When I was running tests for `Chartist.createSvg`, `container.__chartist__` was undefined.

What this does is append the className that is passed to `createSvg` to the `querySelector` for locating the svg element. That way the container may contain other `svg` elements as direct child or children of other nodes.

I added a test for this which required I create a new spec; `spec-core`. I added jQuery to the Jasmine config to make creating and finding DOM elements easier in the specs.

I noticed that `jasmine-fixture` is included in the dependencies, but I could not figure out how to include it in the test runner.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "grunt-umd": "~2.2.1",
     "grunt-usemin": "~2.6.2",
     "handlebars-helpers": "~0.5.8",
-    "jasmine-fixture": "~1.2.2",
+    "jasmine-jquery": "~2.0.5",
     "jshint-stylish": "~1.0.0",
     "load-grunt-config": "^0.16.0",
     "lodash": "^2.4.1",
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -164,12 +164,13 @@ var Chartist = {
     width = width || '100%';
     height = height || '100%';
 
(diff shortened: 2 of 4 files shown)


tylergaw:
I'm testing this fix out on a project I'm working on with ember-cli-chartist: https://github.com/tylergaw/ember-cli-chartist/releases/tag/v0.2.3-beta1 I just included a build of chartist.js from my fork.

So far it's functioning well. Will keep kicking it around to see if there are any issues.

gionkunz:
@tylergaw , thanks for your contribution! This is awesome! Specially that you've included specs :+1: 

I'm still not 100% sure about this as it seems like a hack to use the CSS class of the SVG in the query selector. Consider that the current function call should remove the SVG of a previous call, that could theoretically be called with a different class name.

I'll pull your fork into a branch in my repo and we can work together from there to test this further.

gionkunz:
I've created a new pull request here: https://github.com/gionkunz/chartist-js/pull/147

Can you take a look and also test your use cases?

Cheers
Gion

PR #147, opened 2015-01-15
Tylergaw only remove chartist svg

gionkunz:


tylergaw:
Yep, this is way better. You're right the className way I was doing was hackish. All my test cases are still working. I did a new beta build [ember-cli-chartist (0.2.3-beta2)](https://github.com/tylergaw/ember-cli-chartist/releases/tag/v0.2.3-beta2) with these changes, and it's working as expected. Same thing in a project that's using it.

Also way cooler using jasmine-jquery for the specs.

Big :+1: all around from me.

gionkunz:
Cool thanks for reviewing the PR! I've just merged it back to develop and it will go to the next release I planned to publish on sunday.

Cheers
Gion

PR #149, opened 2015-01-15
Added ability to include metadata for rendered points

akharazia:
This allows for the ability to add metadata to point elements for contextual events / tooltips (showing date, other related values), etc.


gionkunz:
Hey @akharazia thanks a lot for your contribution! I like the idea of including meta data in the series! There are a few problems with your PR though. You should never include dist files in our PR. Also I'd appreciate this functionality to be implemented as plugin http://gionkunz.github.io/chartist-js/plugins.html once the data events are implemented with #154 . This way we are not bloating the core functionality of Chartist too much. I hope you don't mind I close / reject your PR right away. If you feel like implementing this in a plugin once the data event is there. I'd be more than happy to accept it and add it to the list of plugins on the Chartist site!

gionkunz:
I've changed my mind on this. I'll include meta data information in the core. This becomes beneficial for #151 and shouldn't create the overhead of including a plugin. I'll use your PR as inspiration and implement this for today's release.

Issue #92, opened 2014-11-26
Option to make graph smoothing less agressive


danieldiekmeier:
Hey,

I love Chartist and use it on my reading platform [Lesetagebuch](https://lesetagebu.ch/) to show statistics about the finished books and pages per month. But if somebody didn't read something in one of the months, the graph falls beneath the 0-line. That's not really what I want.

![screen shot 2014-11-26 at 13 34 46](https://cloud.githubusercontent.com/assets/2098462/5201353/04269708-7571-11e4-81a1-aa5452d42d0b.png)

Is there a possibility of toning this down a bit?

Regards,
Daniel

gionkunz:
Salü Daniel :-) Thanks for the kind words. Chartist.js currently uses a interpolation technique called catmull rom spline. It's very accurate when it comes to approximation to the data points, however, it can create some artifacts when there are big jumps in the data. I think it would be nice to add other interpolation algorithms like https://en.wikipedia.org/wiki/Cubic_Hermite_spline#Cardinal_spline which are accurate too but create less or different artifacts. There is a lot on our plate currently and we are missing some contribution so unless we can find someone who is taking care of this, I think it would take some time until I can pick this up. In the meanwhile the only thing that comes to my mind is working around this issue by either disabling line smoothing for your line chart or by configuring a low that is below your lowest data point so that the artifacts from catmoll rom is not overflowing.

Besides your issue, may I ask you how you achieve to render a Bar chart on top of your Line chart? :-) Do you create a second chart in the same container and disable grid and labels?

danieldiekmeier:
Hey Gion, I just had a look into the source code, so maybe I'll add some more smoothing methods myself, if I find the time. (But as I have an acute interest in this, that may _actually_ happen.)

Yes, I'm just rendering two charts, and disable everything except the chart and its y-labels on the second one. I'm also translating the line chart by half a step so it overlaps on the right spots. You can have a look into the dev tools, if you want: https://lesetagebu.ch/von/danjel

gionkunz:
Cool :-) We are also working on a compound chart type #37 but I guess that's a nice workaround :-)

maslowivan:
:+1:  waiting for fixed

gionkunz:
[...]

PR #155, opened 2015-01-19
Added parameter for smoothing in Line Chart

danieldiekmeier:
This is related to #92, but the smoothing still isn't quite what I want. (But to have full control, I'd have to set all the bezier handles myself.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ You will need the following to run a local development enviroment.
 1. `cd` into your local copy of the repository.
 2. Run `npm install` to install dependencies located in `package.json`.
 3. Run `bower install` to install bower dependencies.
-5. Run `grunt serve` to start the watch task, and the web server should automatically open. Congrats, you should now be able to see your local copy of the demo site.
+5. Run `grunt preview` to start the watch task, and the web server should automatically open. Congrats, you should now be able to see your local copy of the demo site.
 
 ## <a name="submit"></a> Submission Guidelines
 
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -722,8 +722,9 @@ var Chartist = {
     return delta;
   };
 
-  //http://schepers.cc/getting-to-the-point
-  Chartist.catmullRom2bezier = function (crp, z) {
+  // http://schepers.cc/getting-to-the-point
+  // https://en.wikipedia.org/wiki/Cubic_Hermite_spline#Cardinal_spline
+  Chartist.catmullRom2bezier = function (crp, c, z) {
     var d = [];
     for (var i = 0, iLen = crp.length; iLen - 2 * !z > i; i += 2) {
       var p = [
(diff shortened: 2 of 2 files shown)


gionkunz:
Hey @danieldiekmeier , Thanks so much for this PR! Can't wait to play around with this :-)

danieldiekmeier:
This mainly adds the parameter to the method according to https://en.wikipedia.org/wiki/Cubic_Hermite_spline#Cardinal_spline

If you want to use it, you have yet to add it to the API. (I didn't want to make assumptions on how to do it)

gionkunz:
Hmmm.... I know what you mean. I've tested a bit with big data differences and very close to the bottom of the chart. When reducing the smoothing it somehow starts to look weird and choppy. Probably we need to research more and find a more intelligent bezier smoothing algorithm
giphy

danieldiekmeier:
Okay, so I just tried something else (in Sketch.app):

screen shot 2015-01-30 at 01 23 43

This is what it'd look like if all the handles are just horizontal, with each one reaching half way to the next node. It doesn't look quite as good as the current algorithm, but much better in the edge cases we discussed in #92. Should I have a go at this?

gionkunz:
Hey @danieldiekmeier . I was looking at some smoothing algorithms also from D3 http://www.d3noob.org/2013/01/smoothing-out-lines-in-d3js.html and it looks like they are having the same issue with their cardinal spline smoothing. The problem is with other smoothing methods that the line is not necessarily reaching to the point and thats not okay. Thats why I appreciate catmull rome.

Your simple but cool idea to just use horizontal handles for line smoothing is actually very nice and I'd appreciate an algo like this. We could make it configurable so that one can decide which algo will be used.

D3 also offers a monotone spline http://en.wikipedia.org/wiki/Monotone_cubic_interpolation but I'm not sure if your result is actually better. Also this spline could be handy with its 3 parameters http://en.wikipedia.org/wiki/Kochanek%E2%80%93Bartels_spline . 

It would actually be nice to implement these algorithms and give the choice to the user which he wants to use.

Maybe you can give your method a shot as I think its a special case for a spline that only translates vertically. Maybe by having just vertical handles is perfect for this kind of spline.

martinsznapka:
Morris.js does it in this way:
https://github.com/morrisjs/morris.js/blob/master/lib/morris.line.coffee#L301-L309

For comparison - Chartist.js red, Morris.js blue:
diagrams

gionkunz:
Okay thanks @martinsznapka ! Thats exactly what @danieldiekmeier propsed but instead of half width mirror.js uses quarter width for the control points. @danieldiekeier, ill refactor the code a bit so we can make the line shmoothing function configurable. Can you then implememt your suggestion? Also it might be cool to make the control point length divisor configurable and default to 2.

danieldiekmeier:
Sure thing! Just let me know when you're done with the refactoring.

gionkunz:
I've closed this PR as I pushed the new interpolation module. It uses the new SVG path API. It's factory function with optional options that return functions with input path coordinates and should return an Chartist SVG path object. Let's track further progress here #92

PR #168, opened 2015-01-28
Allow series object to override global line chart settings.

atavistock:
In some cases its valuable to override the global settings for a given series in a line chart.  

In my particular case, I wanted `{ showPoints: true, showLine: true, showArea: false}` for most series but wanted to render a single series as the baseline as a the `{ showPoints: false, showLine: false, showArea: true}`.   

This change allows for a series to look like:

```
      {
        className: "ct-series-baseline",
        name: "baseline data",
        data: [1,2,3,4],
        showArea: true,
        showLine: false,
        showPoint: false
      }
```

### DIFF
--- a/src/scripts/charts/line.js
+++ b/src/scripts/charts/line.js
@@ -169,6 +169,12 @@
     this.data.series.forEach(function(series, seriesIndex) {
       seriesGroups[seriesIndex] = this.svg.elem('g');
 
+      // Allow chart options to be overridden on a per series basis
+      var showLine = (series.showLine !== undefined ? series.showLine : options.showLine),
+          showArea = (series.showArea !== undefined ? series.showArea : options.showArea),
+          showPoint = (series.showPoint !== undefined ? series.showPoint : options.showPoint),
+          lineSmooth = (series.lineSmooth !== undefined ? series.lineSmooth : options.lineSmooth);
+
       // Write attributes to series group element. If series name or meta is undefined the attributes will not be written
       seriesGroups[seriesIndex].attr({
         'series-name': series.name,
@@ -192,7 +198,7 @@
 
         //If we should show points we need to create them now to avoid secondary loop
         // Small offset for Firefox to render squares correctly
-        if (options.showPoint) {
+        if (showPoint) {
           var point = seriesGroups[seriesIndex].elem('line', {
             x1: p.x,
             y1: p.y,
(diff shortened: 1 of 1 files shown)


gionkunz:
Hey there. Thanks for the PR! I'm not 100% sure about this change though. These settings will not work with the responsive settings override and will override whatever is inherited in the settings chain. I can imagine that its required to have different series formatting on the same chart but what, except a bit of a performance overhead, speaks against enabling everything and use CSS with series selectors to `display: none` some stuff? I know it would be easy to have it in JS and use the series objects but its just another special case we'd need to consider and Im really trying hard to keep everything simple.

atavistock:
Heyya - 

So I definitely appreciate the philosophy of making allowing these to be controlled by CSS.  It gives a lot of dynamic flexibility to the chart which is awesome and powerful -- I'm leveraging that in my use case.  I also can see how this change overrides the `responsiveOptions`.

In my case, I'm rendering a graph with eleven lines, but only one (the baseline) has an area.  I'm already toggling via tabs and using `display:none`.  If I need to do this outside of chartist, I can but it will be more confusing (the general setting is `showArea: true`, but in most cases theres no area) and I'll need to keep track which lines have areas in via a few more classes.  I'm not sure what the impact will be of emitting everything where most are hidden.  Also, even if I used `responsiveOptions`, I'd not be overriding the settings on my exception cases.  Still I get it... and can work around it...

With that said, if the philosophy is to try to use CSS, then everything should be emitted in every case (points, lines, area).  Only options that truly change the behavior should remain (for example `lineSmooth`).  This would simplify the javascript in the core of chartist at the expense of more complex CSS and external js.  Even `responsiveOptions` should be done using media queries and CSS, rather than being nested inside of the chartist code, essentially making `responsiveOptions` a non-entity.  Obviously this is a different direction than the existing code, and wouldn't be backwards compatible. 

I'll be using my fork for a while because its what I needed to do to get my project moving forward, but will probably move to the pure CSS idea in a little while.

gionkunz:
Hey @atavistock . I thought about this again. This feature can actually be very handy in some cases I can imagine. What about this style of doing the config?

``` javascript
  labels: [1, 2, 3, 4, 5],
  series: [{
    name: 'one',
    data: [7.2, 6.8, 6, 2.6, 7]
  } ,  {
    name: 'two',
    data: [7, 6.5, 5, 2.3, 3]
  }]
}, {
  fullWidth: true,
  showPoint: false,
  showArea: true,
  series: {
    one: {
      showArea: false,
      lineSmooth: false
    }
  }
}, [
  ['screen and (min-width: 320px)', {
    showPoint: true,
    series: {
      one: {
        showPoint: false
      },
      two: {
        showArea: false
      }
    }
  }]
]);
```

Then we have the ability to configure for individual series, but we don't loose to ability to override in the responsive options. What do you think?

Cheers
Gion

gionkunz:
@atavistock I've implemented this on develop. It will come into the next release. Here's an example:

``` javascript
      labels: ['1', '2', '3', '4', '5', '6', '7', '8'],
      series: [{
        name: 'series-1',
        data: [5, 2, -4, 2, 0, -2, 5, -3]
      }, {
        name: 'series-2',
        data: [4, 3, 5, 3, 1, 3, 6, 4]
      }, {
        name: 'series-3',
        data: [2, 4, 3, 1, 4, 5, 3, 2]
      }]
    }, {
      fullWidth: true,
      series: {
        'series-1': {
          lineSmooth: Chartist.Interpolation.step(),
          showArea: true,
          showLine: false,
          showPoint: false
        },
        'series-2': {
          lineSmooth: Chartist.Interpolation.simple()
        },
        'series-3': {
          showPoint: false
        }
      }
    });
```

PR #190, opened 2015-02-11
Added simple smoothing function

danieldiekmeier:
As we discussed in #92, I added a new smoothing function with my idea from #155.

It can be used like this:

``` javascript
  labels: [1, 2, 3, 4, 5],
  series: [
    [1, 2, 8, 1, 7]
  ]
}, {
  lineSmooth: Chartist.Interpolation.simple({divisor: 2})
});
```

The default divisor is 2, this gives the best results, in my opinion.


danieldiekmeier:
I didn't really know which name was fitting, but I'm not set on `simple`, so choose whatever you want.

danieldiekmeier:
Okay, I just noticed that there might be a problem when activating the area-option. (See screenshot). I'm not sure what's going on there, can you have a look at it, @gionkunz?

screen shot 2015-02-11 at 17 20 57

(Online at https://lesetagebu.ch/von/danjel)

gionkunz:
@danieldiekmeier :+1: Cool thanks for your work! I hope I can work on it this evening.

gionkunz:
I like divisor 3 the most! :-)

PR #200, opened 2015-02-21
Catch failures to get bounding box height / width.

jarib:
Using Chartist 0.7.2 in a small React app gives me a somewhat opaque NS_ERROR_FAILURE and no charts rendered in Firefox 35.0.1

This patch catches that error, and makes the charts render correctly.

I'm not seeing the same problem on the Chartist website, so it could be something in React or my code that is the root cause of this. My project is running [here](http://transcripts-staging.holderdeord.no/) if you want to have a look.

### DIFF
--- a/src/scripts/svg.js
+++ b/src/scripts/svg.js
@@ -314,7 +314,14 @@
    * @return {Number} The elements height in pixels
    */
   function height() {
-    return this._node.clientHeight || Math.round(this._node.getBBox().height) || this._node.parentNode.clientHeight;
+    var h = this._node.clientHeight;
+
+    if (!h) {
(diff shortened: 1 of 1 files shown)


jarib:
If I log the errors in the catch block, this is what it looks like:

image

jarib:
I think my problem was the result of calling `chart.update` before the `created` callback had fired, which worked fine in Chrome but caused some weirdness elsewhere. I've fixed it in my app, feel free to ignore this is if you want to.

gionkunz:
Glad you've found a solution for the problem. This whole width / height detection is very shaky and specially because Chartist is also supported to be initialized inside an SVG container.

PR #201, opened 2015-02-22
add extended 'path' property to data object of pie chart

evansharp:
Cloning the path object into a new property when defining the data object on draw will expose the manipulation methods in Chartist.svg.path which are not currently accessible through 'data.element'; the clone will inherit these extended functions however. This approach requires testing.

### DIFF
--- a/dist/chartist.js
+++ b/dist/chartist.js
@@ -3246,6 +3246,7 @@
           index: i,
           group: seriesGroups[i],
           element: path,
+          path: path.clone(),
           center: center,
           radius: radius,
           startAngle: startAngle,


evansharp:
Tested; the path obj first needs to be extended to expose clone() at this point in the stack.

gionkunz:
Yes, the Pie chart does not yet use Svg.Path :-( However, you can always parse a path from the `d` attribute using `new Svg.Path().parse('M10,10L5,5')` , do your modification and then replace element.attrs.d with `path.stringify()` .

We need to fix the Pie chart asap.

PR #212, opened 2015-03-01
Accessibility plugin THIS BRANCH WILL NOT BE MERGED

gionkunz:


gionkunz:
This branch contains a first version of the accessibility plugin for chartist. The template accessibility.hbs is containing the initialization of a Line, Bar and Pie chart, including the accessibility plugin settings.

This PR is just for review purpose and will not be merged. The plugin will land in its own repository at a later point.

In order to launch the test site just start the stack with `grunt dev` and access the page http://localhost:9000/accessibility.html

@JamieKnight , It would be super awesome if you could review the plugin and see where it could be improved. The current settings / data in the accessibility.hbs JS part are set for German as this is what my NVDA speaks. It would be super cool if you could mess around a bit with it and test it with scenarios you can think of.

@Founts I know this is not based of a table but will generate a Table, but maybe you can review this too.

PR #239, opened 2015-03-17
Update to getting-started.yml

trippingtarballs:
After completing the `bower install chartist --save` step, the instructions for setting up _THE SASS WAY_ includes the wrong bash command. This fix is for the listed path to file `_chartist-settings.scss styles`.


gionkunz:
Thanks!

PR #270, opened 2015-04-13
Build the dist folder

alexBaizeau:
Build the dist folder so that the most recents fixes can be used


gionkunz:
Hi @alexBaizeau Sorry that it took so long but 0.7.4 is out now.

Cheers
Gion

PR #277, opened 2015-04-18
Convert string values to numbers

Globegitter:
The API I am using returns strings rather than numbers (so `series: ["1", "2", "3"]), which for some reason is no issue in chrome, but in firefox it throws some cryptic error on this line: https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/core.js#L339

This fixes it. Tested on Chrome, Safari and Firefox.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -331,6 +331,11 @@ var Chartist = {
       } else {
         array[i] = localData;
       }
+      
+      // if value is a string rather than a number (or object), just convert it implicitly
+      if (typeof(array[i]) === 'string') {
+        array[i] = +array[i]
+      }
 
       // Convert object values to numbers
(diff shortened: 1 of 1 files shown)


gionkunz:
Hey @Globegitter . Thanks a lot for your contribution. While reviewing your PR or better while reviewing the code and behavior that was already there, I found the whole getDataArray too inconsistent and complicated. I decided to refactor the whole code. Nevertheless, your contribution is highly appreciated! Please check the commit 85e4e37 to see if your issues are fixed with these changes too. Please also note that I've added some test specifications to reflect this use case. We should always try to add tests for such scenarios.

Globegitter:
Thank you @gionkunz the fix is highly appreciated and yep will keep adding tests in mind for the next PR :)

PR #284, opened 2015-04-24
Prevent Firefox from dying with a DOM exception when calling getBBox() on an invisible node.

scthi:
Firefox throws a NS_ERROR_FAILURE when calling  getBBox() on an invisible node.
See http://jsfiddle.net/sym3tri/kWWDK/

### DIFF
--- a/src/scripts/svg.js
+++ b/src/scripts/svg.js
@@ -307,6 +307,24 @@
   }
 
   /**
+   * "Save" way to get property value from svg BoundingBox.
+   * This is a workaround. Firefox throws an NS_ERROR_FAILURE error if getBBox() is called on an invisible node.
+   * See [NS_ERROR_FAILURE: Component returned failure code: 0x80004005](http://jsfiddle.net/sym3tri/kWWDK/)
+   *
+   * @memberof Chartist.Svg
+   * @param {SVGElement} node The svg node to
+   * @param {String} prop The property to fetch (ex.: height, width, ...)
+   * @returns {Number} The value of the given bbox property
+   */
+  function getBBoxProperty(node, prop) {
+    try {
+      return node.getBBox()[prop];
+    } catch(e) {}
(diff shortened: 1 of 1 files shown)


src/scripts/svg.js
```diff
@@ -307,6 +307,24 @@
...
+   * "Save" way to get property value from svg BoundingBox.
+   * This is a workaround. Firefox throws an NS_ERROR_FAILURE error if getBBox() is called on an invisible node.
+   * See [NS_ERROR_FAILURE: Component returned failure code: 0x80004005](http://jsfiddle.net/sym3tri/kWWDK/)
+   *
+   * @memberof Chartist.Svg
+   * @param {Node} node The svg node to
+   * @param {String} prop The property to fetch (ex.: height, width, ...)
+   * @returns {Number} The value of the given bbox property
+   */
+  function getBBox(node, prop) {
```
gionkunz: First thanks for the PR! It looks good to me. I also need to do cross browser testing as well as check if this also works in SVG context but I guess it should be fine. It's also kind of an undocumented feature that you can initialize Chartist on a SVG element too.

One thing on this line can we maybe go for the name getBBoxProperty instead of getBBox? Just to eliminate any confusion about the original name and this wrapper with different return type and signature but the same name.

Cheers
Gion

scthi:
I changed the method name to getBBoxProperty() as you suggested.
Chartist is a great lib btw.

Regards
Thilo

gionkunz:
Hi Thilo

Great! Thanks for your contribution! And thanks for the kind words :-) There are millions of things we still need to fix and change but we do our best in the available time that we have in an open source project.

Cheers
Gion

hisnameisjimmy:
It looks like this was merged into develop, but I'm still seeing this in the current dist (0.11.0).

Did this ever get tested?

PR #292, opened 2015-04-26
Added additional check to getHighLow() to support empty charts.

scthi:
Without this check the browser freezes (endless loop in Chartist.getBounds() function) with input data:

```
data = {
      labels: [],
      series: []
    }
```

I know, it would also be possible to check  `dataArray.length === 0`, but I overwrite Chartist.normalizeDataArray:

```
Chartist.normalizeDataArray = function (dataArray) {
  return dataArray;
};
```

(in order to support "incomplete" series.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -435,25 +435,43 @@ var Chartist = {
    *
    * @memberof Chartist.Core
    * @param {Array} dataArray The array that contains the data to be visualized in the chart
+   * @param {Object} options The Object that contains all the optional values for the chart
    * @return {Object} An object that contains the highest and lowest value that will be visualized on the chart.
    */
-  Chartist.getHighLow = function (dataArray) {
+  Chartist.getHighLow = function (dataArray, options) {
     var i,
       j,
       highLow = {
-        high: -Number.MAX_VALUE,
-        low: Number.MAX_VALUE
-      };
+        high: options.high === undefined ? -Number.MAX_VALUE : +options.high,
+        low: options.low === undefined ? Number.MAX_VALUE : +options.low
+      },
+      findHigh = options.high === undefined,
+      findLow = options.low === undefined;
 
     for (i = 0; i < dataArray.length; i++) {
       for (j = 0; j < dataArray[i].length; j++) {
-        if (dataArray[i][j] > highLow.high) {
+        if (findHigh && dataArray[i][j] > highLow.high) {
           highLow.high = dataArray[i][j];
         }
 
-        if (dataArray[i][j] < highLow.low) {
+        if (findLow && dataArray[i][j] < highLow.low) {
           highLow.low = dataArray[i][j];
         }
+      }
+    }
+
+    // If high and low are the same because of misconfiguration or flat data (only the same value) we need
+    // to set the high or low to 0 depending on the polarity
+    if (highLow.high <= highLow.low) {
+      // If both values are 0 we set high to 1
+      if (highLow.low === 0) {
+        highLow.high = 1;
+      } else if (highLow.low < 0) {
+        // If we have the same negative value for the bounds we set bounds.high to 0
+        highLow.high = 0;
+      } else {
+        // If we have the same positive value for the bounds we set bounds.low to 0
+        highLow.low = 0;
       }
     }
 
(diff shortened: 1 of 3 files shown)


gionkunz:
Hey there! Thanks for your contribution! I agree this is something we can include, but maybe it would be better to include it in the getBounds function as we already do checks there to adjust high and low. ~~Also the calculated high / low gets overriden with the user settings in the options, so it would make sens to do this check after the user settings were used to override.~~ This is not true the override happens later on. So maybe we should move this whole logic, together with the no-data check into the getHighLow function?
https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/core.js#L472-L485

gionkunz:
Sorry what I just said makes no sense at all :-) The configuration override happens in the individual charts at the moment. https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/charts/line.js#L106-L109

But maybe we should change getHighLow to accept a second parameter options and that the whole responsibility of figuring out high and low should be done there. This would include the optional usage of the options (if set) as well as the normalization that is currently part of getBounds.

Id rather move https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/charts/line.js#L106-L109 into getHighLow (of course also for bar.js) but also move https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/core.js#L472-L485 into getHighLow

scthi:
getHighLow() should always return a valid interval (low <= high), or?

Adding the options param with the User defined high low is a good idea, i think.

gionkunz:
no it should always be high > low. If they are the same then high should be low + 1 (as an artifical solution to the problem of expanding in the y axis)

gionkunz:
Will you make the changes?

scthi:
Yes.

gionkunz:
Cool =)

scthi:
Done.

src/scripts/core.js
```diff
@@ -435,28 +435,46 @@ var Chartist = {
...
    */
-  Chartist.getHighLow = function (dataArray) {
+  Chartist.getHighLow = function (dataArray, options) {
     var i,
       j,
       highLow = {
-        high: -Number.MAX_VALUE,
-        low: Number.MAX_VALUE
-      };
+        high: typeof options.high === 'number' ? options.high : -Number.MAX_VALUE,
```
gionkunz: Hey this is great! Thanks so much! One thing, which is more of a style thing I'd like to keep in the library. Id like to always try to keep the balance between being strict and more casual about type conversion. This way the API stays flexible. Can we do `options.high !== undefined ? +options.high : -Number.MAX_VALUE` instead? This way we check if the default (which is `undefined`) was changed and if yes we will do an implicit type conversion to number.

I'm sorry that I seem a bit picky about this but I'd really like to keep up this kind of style.

src/scripts/core.js
```diff
@@ -435,28 +435,46 @@ var Chartist = {
...
+        low: typeof options.low === 'number' ? options.low : Number.MAX_VALUE
+      },
+      findHigh = typeof options.high !== 'number',
```
gionkunz: Same here better with `options.high === undefined`

gionkunz:
I'm sorry to keep on poking around in your PR :-) But I just found that maybe the time is right to update this code to a more functional maintainable style? :-) I just hate those ugly old for loops.

What do you think about this?

``` javascript
function reduceToHighLow(highLow, value) {
  if(value instanceof Array) {
    highLow = value.reduce(reduceToHighLow, highLow);
  } else {
    highLow.high = value > highLow.high ? value : highLow.high;
    highLow.low = value < highLow.low ? value : highLow.low;
  }

  return highLow;
}

var highLow = dataArray.reduce(reduceToHighLow, {
  low: Number.MAX_VALUE,
  high: -Number.MAX_VALUE
});
```

I was running some jsperf tests and the native for loop only seems to be 1% faster (with two series of 1000 values) http://jsperf.com/find-high-low-in-array

gionkunz:
Well, Im sorry I was not reading the jsperf results correctly... its acually almost 20 times slower to use the reduce version... :-( we should probably stick to the lame old native for loop then. Sorry for the confusion.

scthi:
Done.

Most of the time I prefer native loops over functional style.
1. I find them more flexible when it comes to skipping loop iterations/canceling the loop.
2. They are easier to read even if (or because) you can't write everything in one line (IMHO).
3. They work well array-ish things (arguments, NodeLists, ...)
(4. I'm not sure about the the hidden GC costs of the functional style in some environments.)

gionkunz:
Thanks @scthi !

gionkunz:
Yeah, there are pros and cons. I personally prefer functional style over traditional procedural for a few reasons. https://mindtheshift.wordpress.com/2014/08/13/functional-style-javascript/

gionkunz:
Merged! Thanks again for your work and patience!

PR #320, opened 2015-05-17
Fixed SVG duplication for old browsers

radist2s:
My commit fixes svg duplication bug for old browsers like Opera 12. Bug occurs when you updated graph.
Upper graph is not removed
(Upper graph is not removed)


gionkunz:
@radist2s do you want to make the change for `getAttributeNS('http://www.w3.org/2000/xmlns/', 'ct')` and file a PR?

radist2s:
I created another PR: https://github.com/gionkunz/chartist-js/pull/352

Issue #306, opened 2015-05-06
Bars overlay labels when drawn in Firefox


alexZielonko:
I'm successfully overlaying horizontal bars with label information in Chrome. However in Firefox the labels are overlayed by bars.

Chrome:
![image](https://cloud.githubusercontent.com/assets/5274697/7505276/9540b6e6-f416-11e4-96ad-039bd97003e6.png)

Firefox:
![image](https://cloud.githubusercontent.com/assets/5274697/7505288/a58ccef4-f416-11e4-8e31-a61db8d26211.png)

Currently the bar chart is being drawn in the order following order: labels -> grids -> series.

I believe it is more logical to draw the chart in this order: grids -> series -> labels. 

Please advise if there is an existing way to resolve my display issue. If not, I'm happy to discuss further and/or make a pull request.

gionkunz:
Hey there. Can you share your code? It's a bit hard to think about this if I can't see how you've implemented it. The thing with the lables is that they are actually part of the grid. They are closely coupled together in Chartist. If the label interpolation function returns falsey the grid line wont be drawn either.

You can use the `draw` event of Chartist to generate your own labels. Maybe this example helps? http://jsbin.com/wacuva/2/edit?css,js,output

Btw. I'm working on getting the events documented...

alexZielonko:
The coupling between label and grid is useful for label interpolation.

While the example you provide is useful and appreciated, my current issue is more closely related to the SVG's painters model of rendering. It seems logical to assume I never want a label to be obscured by data. I can read data under a label, but not necessarily a label under data. When physically painting a chart on a wall, I can't imagine painting the labels prior the grid. Further, if the painter knows the labels will overlay the data, the labels would be painted after the data.

Here was a quick commit to get the charts behaving as I expected in Firefox. The only real change is the order in which Chartist.Svg.elem is invoked during bar chart instantiation. While this issue may seem somewhat philosophical, I believe it better adheres to what is expected by the W3's SVG rendering model.

https://github.com/missedtheboat/chartist-js/commit/b994671fb098c24f95526a4a9ab908c81fd8926c

gionkunz:
You're right, it makes sense the labels are drawn on top of everything, even if they are usually outside of the chartRect. Can you make a pull request on the current source files on the develop branch?

alexZielonko:
Cool, I'll get a pull request put together over the weekend.

gionkunz:
[...]

PR #325, opened 2015-05-20
update svg drawing order of bar chart to: grid -> series -> labels

alexZielonko:
https://github.com/gionkunz/chartist-js/issues/306


dist/chartist.js
```diff
@@ -3525,6 +3530,8 @@ var Chartist = {
     showLabel: true,
     // Label position offset from the standard position which is half distance of the radius. This value can be either positive or negative. Positive values will position the label away from the center.
     labelOffset: 0,
+    // This option can be set to 'inside', 'outside' or 'center'. Positioned with 'inside' the labels will be placed on half the distance of the radius to the border of the Pie by respecting the 'labelOffset'. The 'outside' option will place the labels at the border of the pie and 'center' will place the labels in the absolute center point of the chart. The 'center' option only makes sense in conjunction with the 'labelOffset' option.
+    labelPosition: 'inside',
```
gionkunz: Did you rebase before creating the PR? Why is this change in your PR?

gionkunz: Ahh.. sorry this is the dist file. Please never add changes to dist files in PR's. Can you exclude them and resubmit your PR? Thanks!

gionkunz:
Shouldn't this change also be done to the Line chart?

alexZielonko:
I reverted the 'built project' commit and added the same functionality to the line chart. Let me know if any other changes are necessary.

gionkunz:
Hey there. Sorry for my slow pace but I didn't had a chance to work on chartist lately. I just checkout your branch and tested a bit. One thing I've noticed is that due to moving the grid and label creation to the end, also the order of the events changed. So no first the series (line, point, area) events trigger and then the label and grid events trigger. This is not ideal as I believe grid and label events should trigger first.

Can you check if you could leave the order as before but use the insertFirst parameter in the [Svg.elem function](http://localhost:9000/api-documentation.html#chartistsvg-function-elem)? This way you could just change the order how the group elements are placed into the SVG element which would affect appearance but not the structural sequence how the chart is created.

gionkunz:
e.g. `var labelGroup = this.svg.elem('g', null, options.classNames.labelGroup, true);`

gionkunz:
Actually best would be to have all groups created on top of the create function where the order is how they should appear as groups in the SVG element.

gionkunz:
@missedtheboat I'm working on some changes to make Axis Type user configurable and I was incorporating your suggested drawing order together with this change. Thanks for your contribution, it's really appreciated!

PR #330, opened 2015-05-25
Option for drawing donut charts with filled shapes rather than strokes

cdcarson:
Hi --

I added an option to draw donuts with fills rather than strokes. This grew out of my desire to minimize the complexity of CSS customization in my project (which allows users to define fills and strokes, etc.) For example, I'd like to be able to set `stroke: #fff; stroke-width:3px` and have that work for both pies and donuts. I don't think this is possible with donuts drawn with strokes.

Another possible advantage is that the sizes of the donut and donut hole are calculated by percentage of the total radius -- you don't have take into account the stroke width. 

Here's what I've done:
- In `pie.js`:
  - Left the old "donut" functionality alone: that is, existing donuts won't be affected or break.
  - Added a chart option called `donutHoleFraction`. If this is a number greater than 0 and less than 1 and `donut` is true, the chart will be drawn with a fill method (ignoring `donutWidth`); otherwise everything is drawn as before.
  - Added logic around label placement when we're using the fill method.
  - Added code to draw the filled slice.
  - Slices drawn with this method are given the class `ct-donut-shape` (`options.classNames.donutShape`) rather than `ct-donut`.
- in `site`
  - Added a couple of examples. These work as expected as far as I've been able to test.

There wasn't any need to mess with the stylesheet. 

Let me know what you think.  Thanks!

--Chris


gionkunz:
Thanks so much for the PR and sorry that I couldn't get back earlier. Overall I totally agree with the feature and find it very useful for certain situations (like if you'd wan't to draw an outline on your donut or if you don't want to rely on large stokes that can cause rendering problems). Also the examples and end to end integration into the documentation site is much appreciated :+1: 

About the consistency for API users I'm not 100% sure yet. I'd like to avoid any dead options, as if you enable shape donut by putting `donutHoleFraction` greater than 0, you'll kill `donutWidth`. 

Also I'm sure there are usecases for having a relative unit for specifying the donut width by using the `donutHoleFraction` , I still think it would be nicer for the designers to specify the width of the Donut.

Looking at your code (which is really awesome, thanks again!) I would wish for something like this:

Regular Donut chart using strokes

```
new Chartist.Pie('.ct-chart', {
  series: [20, 10, 30, 40]
}, {
  donut: true,
  donutWidth: 60,
  startAngle: 270,
  total: 200,
  showLabel: true
}); 
```

Solid donut chart using paths

```
new Chartist.Pie('.ct-chart', {
  series: [20, 10, 30, 40]
}, {
  donut: true,
  donutWidth: 60,
  donutSolid: true,
  startAngle: 270,
  total: 200,
  showLabel: true
}); 
```

Then I'd use `donutWidth` too for the calculation of the width of the outer and inner arc rather than a factor.

Also I'd prefer the word "solid" for this feature. Like `donutSolid: 'ct-donut-solid'` and the option `donutSolid: true`.

What do you think? Again, thanks for this sophisticated PR!

gionkunz:
Also make sure you rebase, as I've changed the label positioning mechanics a bit in 0.8.2.

cdcarson:
@gionkunz No worries about timing. Things take time, unfortunately :)

I agree about naming and options -- in particular not having dead or proliferating options.  I want to get your input on donutWidth before proceeding...

Let's say we have: `{donutWidth: 60, donutSolid: true}`.  In this case the width of the donut is an integer pixel value and the donut hole would be the donut radius minus 60px. That's fine.

But there is a case to made for having the donutWidth be a percentage, e.g, if displaying a chart at radically different sizes. If donutWidth = 50px the chart looks substantively different at r = 100px  and r = 300px. If donutWidth = 50% the look stays the same. 

So, can we allow  `donutWidth` to be a a CSS percentage value, e.g., `{donutWidth: '50%'}`? I'd check for this case with a regular expression. If it's a percentage, then we calculate the pixel value for donutWidth from the radius, otherwise we treat the original `donutWidth` as the pixel value.

Let me know what you think about this. If you like it, I can modify my existing changes pretty quickly. Thanks,

Chris

gionkunz:
@cdcarson sorry again for me taking so much time to answer :-)

I agree that sometimes it's very useful to specify `donutWidth` in relative units rather than absolute pixels. However, I somehow don't like the effect on the code as we currently only use one unit everywhere which is pixel specified in numbers. If we'd want to introduce more units, we'd need to create an API very carefully also to not introduce too much complexity. Also, we have the concept of responsive setting overrides http://gionkunz.github.io/chartist-js/getting-started.html#responsive-setting-overrides that we'd like people to use when they want to change configuration on different media. However, I still can see the concern of using relative units especially in the donut chart.

Okay, what about this. Let's add a relative unit there. But lets design the API convenience in a way that it's a bit cleaner in the code than relying on string recognition and regex. Let's just assume someone would never want to specify an absolute unit (pixels) between 0 and 1, and then let's say if someone specifies a value in that range (eg. 0.3333) then we treat it as a relative unit (33.33%). Very similar to your initial fraction option. What do you think? This way we can keep the interface number only and in the code we do a simple `if (donutWidth < 1)` check and then use it to multiply with the radius to determine actual pixel value.

This could be implemented as a simply normalization rule:

``` javascript
var donutWidth = options.donutWidth < 1 ? options.donutWidth * radius : +options.donutWidth;
```

From that point on we simply use donutWidth insetad of options.donutWidth.

What do you think?

gionkunz:
Hi Chris

Any update on this? I'd really like to include this feature and it would definitely provide some value to the library.

Cheers
Gion

gionkunz:
Nothing to hurry, just want to make sure I'm not loosing track of this valuable PR :-)

gionkunz:
@cdcarson Hey there. Since I didn't heard from you for some time now, I assume your priorities have changed and you did not had time to work on this any further. If you agree I can take your efforts so far and continue your work?

richardaiko:
:+1: It'll be awesome if this was pulled in, currently deciding btw c3js and chartist. loving chartist so far, but the lack of this feature might be a deal breaker  :(.

Looking to implement a donut similar to:

screen shot 2015-09-09 at 2 23 41 pm

raikoroje:
Looking to implement a donut similar to:

screen shot 2015-09-09 at 2 23 41 pm

hansmaad:
What's wrong with `donutWidth: '50%'`? We already have percentage values in `height` and `width`. `donutWidth: '50%'` was the first thing I tried before I looked into the docs.

``` javascript
var options = {
   height: '60%',
   width:  '80%',
   donutWidth: 0.5  // hm, strange..
}
```

gionkunz:
I'm fine with that but let's just keep it simple. I really don't want to bloat this thing with Regex's and parsing within the pie chart script. Maybe we can do a quick string test and then use `Chartist.stripUnit` ?

``` javascript
var donutWidth = Chartist.stripUnit(options.donutWidth);

if (typeof options.donutWidth === 'string') {
  donutWidth *= radius / 100;
} 
```

I'd also be fine with a extractUnit function that returns the non-numberic characters in a string within the `Chartist` core so we can use this to test and leave other strings to 

``` javascript
var donutWidth = Chartist.stripUnit(options.donutWidth);

if (Chartist.extractUnit(options.donutWidth) === '%') {
  donutWidth *= radius / 100;
}
```

What do you think?

hansmaad:
What about

``` javascript
function quantity(input) {
  if (typeof input === 'string') {
    var match = (/^(\d+)\s*(.*)$/g).exec(input);
    return {
      value : +match[1],
      unit: match[2] || undefined
    };
  }
  return { value: input };
}

var donutWidth = Chartist.quantity(options.donutWidth);

if (donutWidth.unit === '%') {
  donutWidth.value *= radius / 100;
}
```

gionkunz:
Even better! But we need to make sure to re-use this where ever `stripUnit` has been used so it can be removed afterwards.

hansmaad:
Should I prepare a first PR with these changes? Maybe someone else can then add the `donutSolid` feature.

gionkunz:
That would be great @hansmaad. Thanks for your contribution.

miletbaker:
Hi Guys is there any update on this? i.e. when the merge conflicts will be resolved and a new version released? Thank you.

JoshZA:
:+1:

skovalyov:
So this never got finalized, right? Would really love to be able to add those indents mentioned in #356 :-)

s5b:
I really appreciate the inclusion of percentage in `donutWidth`, thank you _very_ much. But I had to come here to find out that it was available. Doesn't appear to be mentioned in the documentation. The documentation of the `options` appears to be a set of examples, which is never going to be definitive or exhaustive. I recommend an update to the documentation to include a specification of the `options`, as well as the examples. (Or am I missing something?)

http://codepen.io/s5b/pen/YNLGYR?editors=0110

ghost:
+1 Please can you merge in the Filled Shapes, I have modified my local package to include the change and it has been working very well for me ! 

Exactly the desired behaviour for stroke addition and easy config

skovalyov:
I do the same. Published my fork, where I addressed all comments by the maintainer, to our local Sinopia and awaiting for it to be integrated into the main repository and consequently published to npm registry.

anshchauhan:
Can someone please merge this? I need this feature badly because I want strokes in my filled donut.

PR #336, opened 2015-06-02
Fixed area path end vertically

shootaroo:
``` js
new Chartist.Line('.ct-chart', {
  labels: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
  series: [
    [5, 9, 7, 8, 5, 3, 5, 4]
  ]
}, {
  low: 0,
  showArea: true
});
```

before

after

### DIFF
--- a/src/scripts/charts/line.js
+++ b/src/scripts/charts/line.js
@@ -267,7 +267,7 @@
           .move(chartRect.x1, areaBaseProjected)
           .line(pathCoordinates[0], pathCoordinates[1])
           .position(areaPath.pathElements.length)
-          .line(pathCoordinates[pathCoordinates.length - 2], areaBaseProjected);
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there. Thanks for the PR. Can you extend it so it works with any kind of holes in the data? Currently your only handling missing data in the end right?

gionkunz:
@shootaroo I've taken this a bit further by implementing some logic to split a path into segments, filtering for "solid" segments and then drawing the areas based on them. As a side effect there can now be N areas per line chart series. Which is fine I guess, events also get triggered per area. This work I've just committed here: b7ebb60 but thanks again for your efforts and contribution!

shootaroo:
@gionkunz It is more better, thanks!

PR #344, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


gionkunz:
Yeah, why not. I always found it useful when you checkout a bower component, that you can sport the version there. Even if it's not for any use in bower. However, bumping the version there on every release for nothing also sucks :-) Let's get rid of it!

PR #348, opened 2015-06-12
Changed getChartRects width/height calculation to use the correct axis

### LINKED ISSUE #347: chartRect calculation is using wrong axis to figure out minimum width / height
I'm running into a problem when using horizontal bar charts in my application. My `options.axisY.offset` set to 160, and each row in my bar chart is roughly 50 pixels. Because of that, if I want the height of my bar chart to be less than 160, chartist actually stretches the graph to make it 160, which is not desirable.

To give an example, let's say I have a chart with 4 rows, with each row being 50 pixels, for a total chart height of 200 pixels (which is more than our `options.axisY.offset`, 160 pixels):

screen shot 2015-06-11 at 11 45 50 am

If I were to remove a row, the total chart height would be 150 pixels, which is less than the width of the labels on the left side (160 pixels):

screen shot 2015-06-11 at 11 47 34 am

As you can see there's a bit of stretching going on on the right side. If I remove a row again (for a total chart height of 100 pixels) it's a lot more obvious:

screen shot 2015-06-11 at 11 47 09 am

Instead of each row being 50 pixels like I intended, Chartist is stretching the rows to 80 pixels so that the total height stays at 160 pixels. If I remove another row, the height of the last remaining row becomes 160 pixels:

screen shot 2015-06-11 at 11 47 54 am

I've pinpointed the problem to here: https://github.com/gionkunz/chartist-js/blob/45df861eb65f98a00b6c4f2ed4f6a9f37bd20265/src/scripts/core.js#L722

It looks like it's comparing the value of `options.axisY.offset` (plus padding but in our application that padding is 0) with the height of the chart, and picking the greater value.

Would it be possible to add an option to disable this behaviour? What would a solution to this look like?

amsardesai:
Fixes #347 

It seems the `getChartRects` call does not take into account `options.horizontalBar` and compares against the wrong axis. This change just makes sure it compares against the right axis when determining the height of the chart.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -718,8 +718,8 @@ var Chartist = {
     var normalizedPadding = Chartist.normalizePadding(options.chartPadding, fallbackPadding);
 
     // If settings were to small to cope with offset (legacy) and padding, we'll adjust
-    width = Math.max(width, (options.horizontalBar ? yAxisOffset : xAxisOffset) + normalizedPadding.left + normalizedPadding.right);
-    height = Math.max(height, (options.horizontalBar ? xAxisOffset : yAxisOffset) + normalizedPadding.top + normalizedPadding.bottom);
+    width = Math.max(width, yAxisOffset + normalizedPadding.left + normalizedPadding.right);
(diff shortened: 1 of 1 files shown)


gionkunz:
Great thanks!

PR #352, opened 2015-06-16
Fixed SVG duplication for old browsers

radist2s:
Changed commit for [rejected pull request](https://github.com/gionkunz/chartist-js/pull/320).


gionkunz:
That should do the trick! Thanks!

PR #355, opened 2015-06-18
Add support for "low" option to Bar chart.

skoutuk:
Currently bar.js has a hard coded referenceValue = 0 when creating the axis. This prevents positive "low" values from being specified and allowing you to zoom-in on the graph as the documentation suggests.

I have made changes to allow the referenceValue to be overriden by the "low" option if specified, thus allowing the value axis to start from values other than zero.

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -170,7 +170,7 @@
 
       valueAxis = axisX = new Chartist.AutoScaleAxis(Chartist.Axis.units.x, data, chartRect, Chartist.extend({}, options.axisX, {
         highLow: highLow,
-        referenceValue: 0
+        referenceValue: options.low || 0
       }));
     } else {
       labelAxis = axisX = new Chartist.StepAxis(Chartist.Axis.units.x, data, chartRect, {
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there! Thanks for the PR. Can you tell me more about the usecase of this feature? I guess this has never been really working very well for the bar chart. However, your PR does not account all cases I think. The problem is also with the `high` option. If you specify a `high` value below zero you get the same issue. We should rather use `Math.min` and `Math.max` to ensure the reference value is always larger or equal than `low` and always smaller or equal than `high`. I've also noticed that the bars get drawn into the padding of the chart which should probably also be prevented during the drawing process of the bars.

skoutuk:
Hi Gion, this is what we were trying to get working... when all the series values are close together, we wanted to be able to start the value axis a lot closer to the actual values, rather than zero.

``` javascript
var data = {
    labels: ['Audi A3', 'MINI Paceman (13 on)', 'BMW 1-Series', 'Volkswagen Golf'],
    series: [
        [325, 330, 330, 341]
    ]
}

var options = {
    axisX: {
        labelOffset: {
                x: -10
        },
        labelInterpolationFnc: function (value) {
                return value + 'L';
        }
    },
    axisY: {
        offset: 100
    },
    horizontalBars: true,
    low: 315
};

new Chartist.Bar(chart[0], data, options);
```

If a low option isn't specified, I think the axis should work as it currently does and start from zero (even using AutoScaleAxis). I haven't encountered any problems using a high option yet, as long as it is greater than that actual values.

cheers
Rich

gionkunz:
I've fixed some of this stuff in the context of making the bar chart axes dynamic. Can you test again with the dev branch?

skoutuk:
Yep, works for me. Thanks!

PR #361, opened 2015-06-22
Adding Variable Data Lengths & Line Breaks

codebend3r:
- options.allowVariableDataLengths: false by default but can be turned on and chartist will not force values to be 0.
- options.breakLine: disabled by default but can be turned on:

Disabled

```
breakLine: {
  enabled: false,
  limit: null
}
```

As array

```
breakLine: {
  enabled: true,
  limit: [500, 900]
}
```

As number

```
breakLine: {
  enabled: true,
  limit: 500
}
```


.gitignore
```diff
@@ -5,6 +5,7 @@ site/bower_components/
 # build
 .tmp/
 .public/
+.idea/
```
gionkunz, line 8: Ignoring files and folders that get created outside of a project (by IDEs or the system for example) should be part of a global git ignore file on your system https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

dist/chartist.js
```diff
@@ -2536,15 +2539,50 @@ var Chartist = {
...
 
       var pathCoordinates = [];
+      var lastValidValue = false;
+      var hasSmartAlignEnabled = function (series, nextValue, options) {
+        // TODO - test and fix multiple smart aligns
+        // if the current series contains limitData
+        // and the upcoming value is null
+        // and both showLimitLine, breakLine are enabled
+        // and breakLine.smartAlign is also true
+        return angular.isDefined(series.limitData) && nextValue === null && !lastValidValue && options.showLimitLine.enabled && options.breakLine.enabled && options.breakLine.smartAlign;
```
gionkunz, line 2549: Chartist has no dependencies so angular does not exist in this context

gionkunz:
Hi there. Thanks for the PR. There are a few issues with this pull request that I've outlined. Also there are  too many changes in the code for a simple feature or at least I don't get the purpose of those changes.

Can you maybe explain what this PR is about? I don't get the idea behind the changes. What is the `allowVariableDataLengths` option good for? Since data holes were supported in 0.8, Chartist does not fill missing data with 0 anymore http://gionkunz.github.io/chartist-js/examples.html#holes-in-data

Whats the breakLine thing doing?

codebend3r:
Hi Gion, thanks for the feedback. I can submit the changes in smaller
increments if you prefer. I'll take a look at the holes in data thing.

I will have to submit allowVariableLengths first though. This feature
allows the data to be different length. And where it starts and ends will
be dynamically plotted.

I added this to the examples page. If you run grunt dev, you will see what
I mean. I can send you a screen cap later.

codebend3r:
I thought removed that angular dependency.

Anyways, it looks like the holes in data feature is exactly what I was try
to do. So we don't need that anymore.

As for the line break. What I'm attempting to do is pass in one data array
and have it change colour on certain values. But the values have to
increment. Which is not ideal. It looks like this:

new Chartist.Line('.ct-chart',
  {
    labels: ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L',
'M'],
    series: [
      {
        data: [
          1, 52, 85, 150, 223, 351, 589, 714, 981, 1189, 1288, 1390, 1441
        ]
      }
    ]
  },
  {
    allowVariableDataLengths: true,
    showLine: true,
    breakLine: {
      enabled: true,
      limit: [500, 900]
    }
  }
);

So when the value reaches 500 it changes, then 900 it changes again. Does
Chartist currently have this capability?

Chester

gionkunz:
Hi Chester

As for the second feature breakLine. I guess this would better be implemented as a plufin. Did you look into the plugin api yet?

codebend3r:
Hi Gion, I have looked into the plugins section. I will make the changes
and turn it into a plugin.

Btw, I love chartist. You guys or you have done a great job.

Is the data with holes out now?

gionkunz:
@crivas Sorry never got back to you. Yes the data holes feature is out since a while. A plugin would always be highly appreciated! I'll close this PR for now, but anyway thank you so much for your contribution!

PR #402, opened 2015-08-05
Fix typo in hasFeature string #401

### LINKED ISSUE #401: Chartist.Svg.isSupported wrong url?
Hey,

I went through your code and saw how you build your feature-string:

https://github.com/gionkunz/chartist-js/blob/develop/dist/chartist.js#L2187

Just wanted to ask, if this is correct, I have to say that after a bit of googling didn't figure out if it's correct or not and i never used "document.implementation.hasFeature".

Shouldn't it be:

return document.implementation.hasFeature('http://**www.**w3.org/TR/SVG11/feature#' + feature, '1.1');

instead of:

return document.implementation.hasFeature('**www.**http://w3.org/TR/SVG11/feature#' + feature, '1.1');

Thank You

pomm0:
Fix #401

### DIFF
--- a/src/scripts/svg.js
+++ b/src/scripts/svg.js
@@ -533,7 +533,7 @@
    * @return {Boolean} True of false if the feature is supported or not
    */
   Chartist.Svg.isSupported = function(feature) {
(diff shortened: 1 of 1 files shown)


gionkunz:
Awesome thanks for the PR!

PR #404, opened 2015-08-07
IE9 Working by Default

builtwith:
By default responsive charts don't work in IE9 this fixes that issue. Instead of throwing error or trying to use 

``` javascript
window.matchMedia
```

without testing it's availability first.

This is my first ever GitHub contribution I hope I'm doing it right. :heart: :+1:

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -887,7 +887,7 @@ var Chartist = {
       var previousOptions = currentOptions;
       currentOptions = Chartist.extend({}, baseOptions);
 
-      if (responsiveOptions) {
+      if (window.matchMedia && responsiveOptions) {
         for (i = 0; i < responsiveOptions.length; i++) {
           var mql = window.matchMedia(responsiveOptions[i][0]);
           if (mql.matches) {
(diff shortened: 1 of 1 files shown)


tonyjmnz:
I wouldnt remove line `914`, it saved me some debugging time when testing in IE.

gionkunz:
Thanks for the PR! I'll consider this more carefully once I can allocate more Chartist time again!

aKzenT:
Any news on this item?

gionkunz:
Since we've documented to use matchMedia polyfill in IE9, I think we should not apply this change. Sorry that this took so long

estomagordo:
Where is this documentation on how to use matchMedia polyfill in IE9? I know the site contains this reference "For IE9 you need to use a matchMedia polyfill. You should take a look at Paul Irish's matchMedia polyfill." and a link to that Github repository. But it is very unclear to me how this works. How do I utilize Paul Irish's code to make Chartist work in IE9?

It's all very unclear to me, and I fail to see why Chartist wouldn't want for this to work out of the box.

PR #406, opened 2015-08-13
Display the last tick on FixedScaleAxis

tonyjmnz:
When using `Bar` graphs with `Chartist.FixedScaleAxis` and a specified
`divisor` option, the last tick is not displayed. E.g:

With `divisor = 12` and `max(series) = 96`, the calculated graph ticks are
`[0, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80, 88]`

There are indeed 12 ticks, but arguably we don't want to count 0 as a
tick. With the proposed change it would also display 96 at the end of
the axis:

`[0, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96]`

### DIFF
--- a/src/scripts/axes/fixed-scale-axis.js
+++ b/src/scripts/axes/fixed-scale-axis.js
@@ -24,7 +24,7 @@
   function FixedScaleAxis(axisUnit, data, chartRect, options) {
     var highLow = options.highLow || Chartist.getHighLow(data.normalized, options, axisUnit.pos);
     this.divisor = options.divisor || 1;
-    this.ticks = options.ticks || Chartist.times(this.divisor).map(function(value, index) {
+    this.ticks = options.ticks || Chartist.times(this.divisor + 1).map(function(value, index) {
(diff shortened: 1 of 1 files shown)


gionkunz:
I'll test that ASAP! Cheers for the PR!

gionkunz:
Sorry still didn't had time to test this through :-/ Will get to it soon hopefully.

tonyjmnz:
Any news?

IDisposable:
:shipit:

marydavis:
@gionkunz any idea when this will be added or is there a suggested workaround until then? thanks!

marydavis:
the issue is i am using the time series with moment.js line chart so i wasn't calculating my own ticks, just using the divisor. i found a way to workaround it by using this to generate my own ticks instead of relying on the divisor in the meantime :)

dafyk:
This might seem to work but only for short labels (like "96") problem is the last tick has incorectly calculated position and width. In my case all ticks have width 100px, after using this patch last tick has width only 30px.

PR #413, opened 2015-08-21
Added Chartist threshold plugin to documentation site

jmay33:


gionkunz:
I'll merge that locally since I don't want a merge commit! Thanks for pointing out.

PR #435, opened 2015-09-04
Add scale (lin, log) option to AutoScaleAxis

hansmaad:


hansmaad:
We need a check if `if (this.bounds.low * this.bounds.high <= 0)` to avoid an infinite loop. Should we fallback to linear scale or throw an error?

gionkunz:
Fallbacks often causes confusion. I'd prefer an error. What do you think?

hansmaad:
I agree. Any examples how to throw/handle errors in chartist? I'm currently integrating this feature into our application and will add some changes from real life experience :)

gionkunz:
We have no requirement as we tried to keep error handling to a minimum really. We try to be graceful where possible but if not possible we just throw a new Error object: https://github.com/gionkunz/chartist-js/blob/86d34bd4dcc2f9471614606a47e89233d18bf535/src/scripts/axes/axis.js#L110

No sub-classing or anything alike.

gionkunz:
https://github.com/gionkunz/chartist-js/blob/fbf1f37dc360f668025322134c1ac8a933e280c2/dist/chartist.js#L687

hansmaad:
I've added `if (highLow.low * highLow.high <= 0)  throw new Error('...');`
There is still an issue, when `Chartist.getHighLow` returns `low == 0` because all values are equal (`if (highLow.high <= highLow.low)`) . Any ideas how to handle this?

Maybe we can change `getHighLow` to add "real" low and high values to returned object?

hansmaad:
I'm using this implementation in production for a while now and it works very well. But I think the current implementation is more a hack, than a clean integration into chartist code base. I'm wondering how we could improve.
- Deeper integration of scale option (Chartist.getBounds...)
- Create a new axis and put the hack out of chartist. We could even deploy this as a "plugin" or seperate module, since axes are created like `options.axisX.type.call(Chartist,...`, right?

jacobq:
I've been reviewing your implementation of log-scaled axis today and have a few thoughts / questions I'd like to raise for discussion:
- Why is this functionality specific to `AutoScaleAxis`? It seems to me that one might just as well want to have a `FixedScaleAxis` with non-linear proportions. 
- Instead of adding a specially-formatted string as the `scale` property, wouldn't it be simpler to just allow the user to supply a mapping/translation function? Chartist could just implement a default `(x) => { return x; }` linear/identity function, which the user could override. This seems in keeping with the goal of being lightweight. It also avoids constraining the user to the typical cases of log2 and log10; they could easily use natural logarithm instead, for example.
  mapping-functions
- You mentioned that you have some concerns with this experimental implementation (i.e. that it is a "hack" rather than a good solution). Would you mind elaborating a bit more about what you'd like to see improved? (For example, does it feel like this code doesn't belong here? Is this code difficult to test? etc.)

hansmaad:
@jacobq Scaling is about a) bounds and b) projection. I think it's not that easy to supply a single function that could define the scaling. But I like the idea to offer such extensibility.

I don't like my current implementation, because for linear axes we have `Chartist.getBounds` and for log axes it's hacked into `AutoScaleAxis`. As you have pointed out, it's directly coupled to `AutoScaleAxis` and doesn't offer any extensibility. This is just the code I use in my application, to hack a log axis into Chartist.

Maybe it's a good idea to redesign the axes to offer more extensibility, so that it's easy to add some custom bounds / projection functions or use the predefine linearBounds / linearProjection or logBounds / logProjection.

Issue #276, opened 2015-04-16
Stacked bars do not play nicely with bipolarity.


livtanong:
It looks like the issue is caused by the algorithm responsible for positioning the stacked bars. It's getting the base (or bottom, if the value is positive) position of series n from the extremity (or top, if the value is positive) of series n-1. It should clamp to 0.

Note the image below. The first bar actually has a hidden red bar behind it. The bottom of the peach bar should be anchored to the origin, rather than the extremity of the hidden red bar.

![screen shot 2015-04-17 at 03 34 32](https://cloud.githubusercontent.com/assets/869638/7189624/b591e51e-e4b2-11e4-8885-b7fd288bdb26.png)

gionkunz:
HI @levitanong can you share the code to reproduce the issue?

Cheers
Gion

gionkunz:
Also what behaviour would you expect from a stacked bar chart with negative values? I guess a bi-polar stacked bar chart is quite difficult to visualize? Never thought about this to be honest. I think I tried once with only negative values, but not mixed.

gionkunz:
Hey there, I'll close this issue for now. If you can show me a valid use case where you'd include positive and negative values in a stacked bar chart, please comment on further and we can discuss.

moneytree-doug:
@gionkunz With negative values, I would expect this type of chart: http://jsfiddle.net/gh/get/jquery/1.9.1/highslide-http://www.highcharts.com/demo/bar-negative-stack

Is there anyway to achieve that with today's build?

moneytree-doug:
@gionkunz Okay, so I can have it behave that way if I have the stack bars draw from the zeroPoint line at this line: https://github.com/gionkunz/chartist-js/blob/dd3fe00a6b72f9dcddb43a91f39210bbaa73d47e/src/scripts/charts/bar.js#L325

Would it be better to add an option? 
```
stackBars: true,
negativeStack: true  # this would imply that stackBars is true
```

PR #436, opened 2015-09-07
Added an option to allow stacked bars to draw from the zero line

moneytree-doug:
This is related to issue: https://github.com/gionkunz/chartist-js/issues/276

screen shot 2015-09-07 at 4 00 22 pm

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -78,9 +78,9 @@
     seriesBarDistance: 15,
     // If set to true this property will cause the series bars to be stacked. The default stack mode is `accumulate`.
     stackBars: false,  
-    // If set to 'accumulate' this property will force the stacked bars to draw from the zero line.
-    // If set to 'overlap' this property will form a total for each series point. This will also influence the y-axis and the overall bounds of the chart. In stacked mode the seriesBarDistance property will have no effect.
-    stackMode: 'overlap',
+    // If set to 'overlap' this property will force the stacked bars to draw from the zero line.
+    // If set to 'accumulate' this property will form a total for each series point. This will also influence the y-axis and the overall bounds of the chart. In stacked mode the seriesBarDistance property will have no effect.
+    stackMode: 'accumulate',
     // Inverts the axes of the bar chart in order to draw a horizontal bar chart. Be aware that you also need to invert your axis settings as the Y Axis will now display the labels and the X Axis the values.
     horizontalBars: false,
     // If set to true then each bar will represent a series and the data array is expected to be a one dimensional array of data values rather than a series array of series. This is useful if the bar chart should represent a profile rather than some data over time.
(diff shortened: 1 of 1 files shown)


src/scripts/charts/bar.js
```diff
@@ -77,7 +77,9 @@
     // Specify the distance in pixel of bars in a group
     seriesBarDistance: 15,
     // If set to true this property will cause the series bars to be stacked and form a total for each series point. This will also influence the y-axis and the overall bounds of the chart. In stacked mode the seriesBarDistance property will have no effect.
-    stackBars: false,
+    stackBars: false,  
+    // If set to true this property will cause the series bars to be stacked and force all bars to draw from the zero line.
+    negativeStackBars: false,
```
gionkunz: Hey there! Thanks for this PR! I think I like this feature. Can we change the naming, so that we introduce a `stackMode` property that is defaulted to `accumulate` and can be changed to `overlap` ?

moneytree-doug: @gionkunz Yeah, of course! Let me update the PR.

moneytree-doug: @gionkunz Upon writing the comments, I wasn't sure which was which. I am assuming ``accumulate`` would be the behaviour in the screenshot of this PR? And overlap is the existing functionality in master, correct?

moneytree-doug:
@gionkunz Updated. Here are screenshots: 

screen shot 2015-09-15 at 12 55 58 pm
screen shot 2015-09-15 at 12 55 51 pm

gionkunz:
Awesome! Thanks for the work @moneytree-doug !

gionkunz:
Actually I think we were thinking about those names differently :-)

Let me show you what I mean with an image:
img_20150928_112455

With Accumulate, I would like to address the standard behaviour so bars are stacked on top of each other and building a sum in total. 

The overlap mode should provide a way to start all bars from the zero point (reference point) which will be helpful if we style one series bar width smaller so we can see the bar in the background but also when we use one series for negative values and one for positive like in your usecase.

I think currently you've the modes the otherway around, or am I missing something?

Cheers
Gion

moneytree-doug:
@gionkunz You're right. That makes sense. I went ahead and reversed it.

gionkunz:
Fantastic! Thanks for your work, it's really appreciated!

src/scripts/charts/bar.js
```diff
@@ -320,9 +323,20 @@
         var positions = {};
         positions[labelAxis.units.pos + '1'] = projected[labelAxis.units.pos];
         positions[labelAxis.units.pos + '2'] = projected[labelAxis.units.pos];
-        // If bars are stacked we use the stackedBarValues reference and otherwise base all bars off the zero line
-        positions[labelAxis.counterUnits.pos + '1'] = options.stackBars ? previousStack : zeroPoint;
-        positions[labelAxis.counterUnits.pos + '2'] = options.stackBars ? stackedBarValues[valueIndex] : projected[labelAxis.counterUnits.pos];
+        
+        if(options.stackBars && (options.stackMode == 'accumulate' || !options.stackMode)) {
```
gionkunz: One little thing before we merge the PR. Can you change to strict equal here?

moneytree-doug: @gionkunz Done! :+1:

gionkunz:
Awesome! That PR just went in. Thanks again for your contribution!

moneytree-doug:
@gionkunz Thank you for such an awesome library!

moneytree-doug:
@gionkunz Which PR just went in? Should I close this?

gionkunz:
Looks like Ive missed to press the button sorry!

Ice101781:
Thanks so much for adding this feature. One problem I'm having is that, in 'accumulate' mode, the chart won't display correctly if any of the array values in the series equal zero. This feature would be useful to me for a stacked bar chart I'm working with that's based on sentiment. Is there a quick fix for this?

nullablebool:
Great feature, another issue seems to be when using 'overlap' is that the axis dont reflect the data.

The problem being that the axis assumes the values 'accumulate' resulting in using the sum, as opposed to the min and max values of the set.

mddanishansari:
This is my data
```javascript
series: [
  [-150017,  -150017,  -150017,  -150017,  -150017],
  [0,  0,  0,  0,  150034],
  [73389,  152612,  238067,  329380,  426954]
 ]
```

When I use `stackMode: overlap` this is what I'm getting
chart-issue-overlap
Note that `Guaranteed Benefit` bar is getting overlapped by `Non-guaranteed benefit` bar

And when I use `stackMode: accumulate` or just leave it  for default this is what I'm getting
chart-issue-accumulate
Note that in 1st stacked bar positive value is going below negative value and other weird issues like in 3rd stack bar I can see `Premium` bar peeking from behind the `Non-guaranteed benefit` bar 😅

How do I solve this issue? Any kind of help is appreciated

Thank you

PR #452, opened 2015-09-14
Add chartist plugin barlabels example to documentation

tirdadc:
I ended up writing a basic plugin for this functionality, I've updated the documentation with an example and a link to it. Please let me know if I missed anything.


gionkunz:
Hi there. Thanks for your contribution and sorry that I'm coming back to you so late. I was creating something that works for both bars and lines. http://jsbin.com/repesu/1/edit?js,output 

If you update your plugin to handle both and probably an option to rotate the labels, we can remove the current sample plugin and replace it with yours. Can you do these changes?

Cheers
Gion

gionkunz:
Closing becauase of inactivity

PR #466, opened 2015-09-29
Added Chartist fill donut plugin to documentation site

moxx:
a small plugin for animated donuts, it will draw the donut a second time and give him a background/fill look. Also provides the possibility to add multiple html/text labels on different positions.


gionkunz:
Awesome! Thanks for your contribution.

PR #471, opened 2015-09-30
Allow relative donutWidth in percent

hansmaad:
As dicussed in #330
- Added pie chart unit tests.
- Implemented Chartist.quantity, replacing stripUnit
- Changed Chartist.Pie.chreateChart to calculate radius and stroke-width from donutWidth.value

### DIFF
--- a/src/scripts/charts/pie.js
+++ b/src/scripts/charts/pie.js
@@ -192,7 +192,7 @@
       // If this is a donut, we add the stroke-width as style attribute
       if(options.donut) {
         pathElement.attr({
-          'style': 'stroke-width: ' + (donutWidth.value) + 'px'
+          'style': 'stroke-width: ' + donutWidth.value + 'px'
         });
       }
 
--- a/src/scripts/charts/pie.js
+++ b/src/scripts/charts/pie.js
@@ -35,6 +35,7 @@
     // If specified the donut CSS classes will be used and strokes will be drawn instead of pie slices.
     donut: false,
     // Specify the donut stroke width, currently done in javascript for convenience. May move to CSS styles in the future.
+    // This option can be set as number or string to specify a relative width (i.e. 100 or '30%'). 
     donutWidth: 60,
     // If a label should be shown or not
     showLabel: true,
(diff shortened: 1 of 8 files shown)


src/scripts/charts/pie.js
```diff
@@ -186,7 +192,7 @@
       // If this is a donut, we add the stroke-width as style attribute
       if(options.donut) {
         pathElement.attr({
-          'style': 'stroke-width: ' + (+options.donutWidth) + 'px'
+          'style': 'stroke-width: ' + (donutWidth.value) + 'px'
```
gionkunz: do we need the brackets if we got rid of the coercion?

test/spec/spec-core.js
```diff
@@ -279,4 +279,27 @@ describe('Chartist core', function() {
...
+    });
+    
+    it('should return value and unit from string', function() {
+      expect(Chartist.quantity('100%')).toEqual({ value: 100, unit :'%' });
+      expect(Chartist.quantity('100 %')).toEqual({ value: 100, unit :'%' });
+      expect(Chartist.quantity('0px')).toEqual({ value: 0, unit: 'px' });
+    });
+    
+  });
 });
```
gionkunz, line 305: Awesome thanks for adding tests!

test/spec/spec-pie-chart.js
```diff
@@ -33,4 +33,208 @@ describe('Pie chart tests', function() {
...
+     onCreated(function() {
+        $('.ct-slice-donut').each(function() {          
+          var style = $(this).attr('style');
+          expect(style).toMatch('stroke-width:\\s?50px');
+        });
+        done();
+      });    
+    });  
+  });
+
```
gionkunz, line 239: Hey that's fantastic :+1: Your efforts are much appreciated!

src/scripts/core.js
```diff
@@ -79,11 +79,7 @@ var Chartist = {
    * @return {Number} Returns the string as number or NaN if the passed length could not be converted to pixel
    */
   Chartist.stripUnit = function(value) {
```
gionkunz: Can't we get rid of this function? If we don't use it anymore we should remove it.

hansmaad: I didn't remove it because I didn't know if it's part of the public API. It's in the docs at least. Should we mark it as deprecated or just remove it?

src/scripts/charts/pie.js
hansmaad: done :)

src/scripts/core.js
gionkunz: Let's remove it. The docs are currently showing all functions even functions that are used internally. We should include a tag `@internal` or something but then also include notes on the documentation site. Also currently we are running the philosophy that all functions are public to use in the API which will probably strike us sometimes when we refactor code but I think that's fine as long as we address it with our versions and in the changelog.

PR #492, opened 2015-10-26
Returning 1 instead of the value of NaN

adamclifford:
When displaying a line chart with one data point the initial position cannot be determined. The projectValue function returns a NaN which the svg cannot be drawn correctly as NaN is not valid for the x1,x2 values on the svg line object.

### DIFF
--- a/src/scripts/axes/step-axis.js
+++ b/src/scripts/axes/step-axis.js
@@ -28,7 +28,9 @@
   }
 
   function projectValue(value, index) {
-    return this.stepLength * index;
+    var val = this.stepLength * index;
+    //check to see if isNaN;
+    return val !== val ? 1 : val;
   }
 
   Chartist.StepAxis = Chartist.Axis.extend({
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -317,7 +317,9 @@ var Chartist = {
    * @param data
    */
   Chartist.reverseData = function(data) {
+    if(data.labels){
     data.labels.reverse();
+    }
     data.series.reverse();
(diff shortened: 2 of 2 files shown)


src/scripts/axes/step-axis.js
```diff
@@ -28,7 +28,9 @@
   }
 
   function projectValue(value, index) {
-    return this.stepLength * index;
+    var val = this.stepLength * index;
+    //check to see if isNaN;
+    return val !== val ? 1 : val;
```
gionkunz, line 33: Shouldn't we use isNaN as we are sure the result will always be of type Number? I guess `window.isNaN` should only be avioided if you're not sure if the value is a number. Also I think we should return 0 instead of 1.

gionkunz:
Thanks for the PR! I added some inline comments.

src/scripts/axes/step-axis.js
adamclifford, line 33: I don't think so, as it is used as a divisor elsewhere to help determine the step distance between point, leaving it as a  1 (or unity if you are getting technical) stops a divide by 0 situation. I thought that the first time i changed it on our local code base,

gionkunz, line 33: I meant shouldn't we use `isNaN` instead of `val !== val` so it will be:

``` javascript
return isNaN(val) ? 1 : val;
```

It makes the code more explicit and developers who don't know that NaN values compared always result in false (even when compared with themselves) don't get confused.

gionkunz:
I think this is no longer needed as of 0.9.8 labels will be normalized

PR #540, opened 2015-12-11
fire beforeDraw event to allow adjustment of positionalData

benib:
This fires a beforeDraw event just before the grid gets drawn. 

This allows to change the positionalData to have grid lines that overlap the base lines on x and y axis e.g.

The positionalData object nees to be send with the event payload (without going through Chartist.extend) to not loose reference.

Not sure if this fits within the concept, but it is very useful for my usecase.

Let me know what you think and if I should change something.

example:

```
chart.on('beforeDraw', function(data) {
  if (data.type === 'grid') {
    if (data.axis.counterUnits.dir === "vertical") {
      data.positionalData.y1 = data.positionalData.y1 - 5;
      data.positionalData.y2 = data.positionalData.y2 + 5;
    } else if (data.axis.counterUnits.dir === "horizontal") {
      data.positionalData.x1 = data.positionalData.x1 - 5;
      data.positionalData.x2 = data.positionalData.x2 + 5;
    }
  }
}
```

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -789,6 +789,17 @@ var Chartist = {
     positionalData[axis.counterUnits.pos + '1'] = offset;
     positionalData[axis.counterUnits.pos + '2'] = offset + length;
 
+    // Event for grid beforeDraw
+    eventEmitter.emit('beforeDraw',
+      {
+        type: 'grid',
+        axis: axis,
+        index: index,
+        group: group,
+        positionalData: positionalData
+      }
+    );
+
     var gridElement = group.elem('line', positionalData, classes.join(' '));
 
     // Event for grid draw


benib:
The same thing can be achieved by setting attributes on `data.element._node` withing the callback of the `draw` event, but being able to change the size/position of the grid lines before they get added to the svg seems more logical to me.

gionkunz:
Hey Beni :-) Thanks for the PR! Maybe we should think about before events for every draw event? Probably we should add this to 1.0 and have the existing events renamed to after... and add new before... events.

benib:
Yes that would make sense. What I don't like about the backwards compatible solution is that in order to be able to change the positionalData, the callback parameter structure is not exactly the same for the before and after event.

If this could get changed for 1.0, it would be way more consistent.

Not sure about your plans on how to approach 1.0. Let me know if you are interested in a PR for this and against what branch I should develop it.

gionkunz:
Currently I'm just collecting issues for 1.0 but in Q2 2016, I'll start with a branch for 1.0 and we can start working. I'm creating all issues for 1.0 against the milestone in github so we keep track. I've created #579 for the `beforeDraw` and `afterDraw` changes.

gionkunz:
Will be incorporated in #810

PR #544, opened 2015-12-12
Fixed misspelling in overlapping bars example

jaredpetersen:
The original overlapping bars example spelled the month of May as "Mai".


gionkunz:
Haha thanks! :-)

Issue #527, opened 2015-11-27
donut chart not rendered correctly if ratio between first and second value is very small


samuelcambien:
If the ratio between the first and the second value for a donut chart data is smaller than 0,056%, the donut chart isn't rendered correctly. This seems related to issue https://github.com/gionkunz/chartist-js/issues/265, where a similar problem occured when one of the data values was zero. That issue is fixed by now, but the problem remains in case the data ratio is too small.
![screen shot 2015-11-27 at 12 03 27](https://cloud.githubusercontent.com/assets/11772282/11439957/e7a2d1ec-94fe-11e5-9988-176dbc9d2edb.png) ![screen shot 2015-11-27 at 13 44 02](https://cloud.githubusercontent.com/assets/11772282/11441507/fb98169a-950c-11e5-928c-ad951f2da6a3.png)

hansmaad:
@gionkunz I'm working on this. Do you remember the reason for the modification of startAngle here

```
      var start = Chartist.polarToCartesian(center.x, center.y, radius, 
          startAngle - (i === 0 || hasSingleValInSeries ? 0 : 0.2));
```

0.2 leads to negative startAngle. All tests pass and all examples look fine if I comment out this part. Do we really need this?

fix / test: https://github.com/hansmaad/chartist-js/commit/d4fe5100ca6695a26a71d140a8af80881e0e1606

gionkunz:
I think this is coming from a fix when data with 0 was passed in. Can you check the git history? I'll check later and post my findings.

hansmaad:
@gionkunz It's part of the first commit :-D
https://github.com/gionkunz/chartist-js/commit/9db458efe1122885434f010bfef792244cc2584b#diff-96ca44095a13ebcf6b90d06c5e6316e2R68

gionkunz:
:-) Hmmm... Yeah, and now I'm recollecting why it's there. The problem is with that if you remove this, you notice a gap between the slices because of rounding issues. The `0.2` provides a little overlap so there is no gap. This is a common problem in vector graphics and also a practice when drawing vector images. It will be a challenge to write tests for this :-/. We should add sufficient comments. What if we limit with `Math.max(0, startAngle)`?

gionkunz:
I guess this could then be simplified to:

``` javascript
  var start = Chartist.polarToCartesian(center.x, center.y, radius, Math.max(0, startAngle - 0.2));
```

hansmaad:
[...]

PR #566, opened 2016-01-06
Fix donut chart if ratio between first and second value is very small

hansmaad:
Fixes https://github.com/gionkunz/chartist-js/issues/527 "donut chart not rendered correctly if ratio between first and second value is very small"


test/.jshintrc
```diff
@@ -26,9 +26,11 @@
     "beforeEach": false,
     "browser": false,
     "describe": false,
+    "fdescribe": false,
```
gionkunz, line 29: Why do we need those?

gionkunz:
Hi Hannes, thanks for the awesome PR!

test/.jshintrc
hansmaad, line 29: To focus on a single `describe` or `it`. Helpful if you're working on a single feature and you don't want to run thousands of unrelated tests.

gionkunz, line 29: Ah, so you change it back to describe and it if you're done with coding. Cool, then that makes sense.

gionkunz:
You're becoming MVC :-) Most valuable contributor ;-)

Issue #32, opened 2014-09-09
multiple series labels and identifying


vlasvlasvlas:
Hi!! Nice and clean charts!!

i was wondering, i'm using multiple linechart series on my data and i dunno how to identify each line
for each line is a diferent series, how can i label each line / colorline?

Data Example:

```
var data = {

    //my labels
    labels: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri'],

    //my multiple series
    series: [
        //series1: label example: apples
        [5, 2, 4, 2, 0],
        //series1: label example: lemons
        [8, 1, 14, 12, 10],
    ]
};
```

all i got is:
http://oi60.tinypic.com/6q8fhv.jpg

but how do i identify each series?

thanks!!

gionkunz:
Hi there and thanks! :-) They get a default class that is generated. ct-series-a, ct-series-b, ct-series-c etc. You can also override by using the object series notation:

``` javascript
var data = {

    //my labels
    labels: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri'],

    //my multiple series
    series: [
        //series1: label example: apples
        {
          className: 'apples',
          name: 'My nice apples',
          data: [5, 2, 4, 2, 0],
        },
        {
        //series1: label example: lemons
          className: 'lemons',
          name: 'My nice lemons',
          data: [8, 1, 14, 12, 10],
        }
    ]
};
```

Then you'll have two things available, first a class will be set on a group above the form elements (series group) as well as some custom attributes on your element containing the original data values. You can also check the new advanced example to attach behavior: http://gionkunz.github.io/chartist-js/getting-started.html#adding-behavior-to-your-charts

vlasvlasvlas:
woow thanks alot! :)

gionkunz:
You're welcome!

vlasvlasvlas:
and how can i force to use specific color for each line (lemons = yellow line; apples: red line)? thanks!

gionkunz:
You need to use css. Check the chartist css to see what styles are applied. I recommend to inspect your chart in the browser dev tools. You can use various style attributes for svg. Chack this article by Jakob Jenkov which is a great summary http://tutorials.jenkov.com/svg/svg-and-css.html

dominicrico:
You can add them via the css `.ct-chart .ct-series.ct-series-a .ct-line` and `.ct-chart .ct-series.ct-series-a .ct-point` just change the stroke to the color you
[...]

Issue #565, opened 2016-01-05
Enhance documentation system for API docs generation


gionkunz:
There are certain limitations with the current documentation generation
- No proper way to search the API
- Poor structuring sometimes
- No way to document events
- Cross links

We should enhance the API generation using Handlebars, Assemble and the JSDoc parser, so we can get rid of those limitations.

PR #572, opened 2016-01-08
add options.customClassNameMap for adding custom classes to Line Charts

morleyzhi:
Hi!

I needed to set custom colors for lines in a line chart (to (a) create a legend outside of Chartist with the line color, and (b) to have related lines share colors when the chart order was deterministic). I also noticed that some people (like in #32) wanted to set line colors in Javascript.

So I created a line chart option `options.customClassNameMap`, which is a list of strings (or null values). When making the graph, if a dataset's index corresponds with an entry in `customClassNameMap`, it would append that string as a class to the SVG element.

Let me know if you think the interface could use some improvement, or you have any other suggestions! Thanks!


gionkunz:
Hi there. I appreciate your contribution, but I wonder that you missed the point of the series object being conveyer of the class name for the series if set. This is exactly where you've put your code in https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/charts/line.js#L163

The feature is also documented here: http://gionkunz.github.io/chartist-js/api-documentation.html#chartistpie-function-pie but not on the line and bar chart where, what I call object series notation, is also supported. We really lack in documentation and should work on #565  to make this stuff more explicit and easy to find.

``` javascript
new Chartist.Line('.ct-chart', {
  labels: ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday'],
  series: [
    {data: [12, 9, 7, 8, 5], className: 'first-series'},
    {data: [2, 1, 3.5, 7, 3], className: 'second-series'},
    {data: [1, 3, 4, 5, 6], className: 'third-series'}
  ]
}, {
  fullWidth: true,
  chartPadding: {
    right: 40
  }
});
```

Anyways, your contribution attitude is highly appreciated and I'd love to see more PR's like this!

morleyzhi:
Ah great! I saw that documentation for Pie graphs but thought it didn't apply to any other graph types. In that case, I'll close this PR and maybe make another one for documenting this feature more explicitly. Thanks!

PR #575, opened 2016-01-09
Allow empty values to be ignored on pie charts

goynang:
Ignoring empty values in pie charts avoids labels being drawn unnecessarily. This is useful as labels for empty slices can overlap other labels leading to a messy chart.

I've added a configuration option called 'ignoreEmptyValues' that defaults to backwards compatible situation (i.e. false). However, if you change this to 'true' then empty value slices will no longer be rendered (avoiding labelling problems).

Tests included.

May be possible to achieve this via custom labelInterpolationFnc option but couldn't see how so came up with this approach instead.

Other option of course is to avoid having the values in the data in the first place but in my particular use case this is problematic.

PS. This is my first pull request on GitHub so apologies in advance if not useful/appropriate. Thanks for a great charting library!


gionkunz:
Hey there! Thanks for your PR! There seems to be a formatting issue. Can you fix that so the change is more explicit?

gionkunz:
you can use the `.editorconfig` and `.jshintrc` within your editor.

PR #578, opened 2016-01-10
Allow empty pie chart values to be ignored

goynang:
Empty pie chart values can render with overlapping labels. This adds a
configuration option to allow them to be ignored when rendering the pie
chart.

Tests included.

This replaces previous pull request which wrapped an IF block around most of the rendering code (resetting it one level deeper and making the PR unclear). This version avoids that in favour of a simple loop “continue” when empty value met and option set to ignore them.


gionkunz:
Great, thanks!

PR #590, opened 2016-01-21
Polyfill for Function.prototype.bind

ger-oss:
PhantomJS 1.9.7 fails because Function.prototype.bind is not defined: https://github.com/ariya/phantomjs/issues/10522

The polyfill function should be good workaround for that: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Function/bind

PhantomJS is nice for server-side rendering. HTML can be easily converted to a SVG file.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -9,7 +9,29 @@ var Chartist = {
 
 (function (window, document, Chartist) {
   'use strict';
+  // Polyfill for Function.prototype.bind: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Function/bind
+  if (!Function.prototype.bind) {
+    Function.prototype.bind = function(oThis) {
+      if (typeof this !== 'function') {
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there. Polyfills should be included where required and not within the library itself. We also use es6-shim within our jasmine test fixture in order to have all array extras and function.bind but this should go as first script included within your project and not into Chartist.

ger-oss:
Hi, 
didn't see that. 
Could you show me where you defined those shims?

gionkunz:
Sorry I was confusing this with an other project. `grunt-contrib-jasmin` is tasking care of hooking the ES5 Shim in if you use the default template https://github.com/gruntjs/grunt-contrib-jasmine/blob/415ae2cbe5174ccaaf3fb2630ffb9f9a8dd80ebd/tasks/lib/jasmine.js#L68-L82

If you're using Chartist within a PhantomJS environment < 2.0 make sure you load necessary polyfills before you execute any other scripts.

PR #591, opened 2016-01-21
Style hooks providing colors of series for user-built legend

gerbus:
In order for anyone to build a working legend out of the box (i.e. without customizing any scss), rules need to be surfaced that apply the proper "color: " rule. 

Possibly the .ct-legend class is overkill, but I wasn't sure about simply adding the color rule to the .ct-series-# classes.

### DIFF
--- a/src/styles/chartist.scss
+++ b/src/styles/chartist.scss
@@ -112,6 +112,10 @@
   }
 }
 
+@mixin ct-legend-series-color($color) {
+  color: $color;
+}
+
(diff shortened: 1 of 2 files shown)


seeroush:
This would be extremely helpful. It is pretty crippling to have graphs without being able to easily build legends. It is a pretty big turnoff for using the library if we can't build legends.

gerbus:
Definitely.

With the above pull-request, it's as simple as:

`<style>`
`.ct-legend li {`
`list-style-type: none;`
`font-size: 10px;`
`}`
`.ct-legend li span:before {`
`content: '\2022';`
`font-size: 54px;`
`vertical-align: middle;`
`line-height: 10px;`
`margin-right: 5px;`
`}`
`</style>`

`<ul class="ct-legend" id="legend-0">`
`<li><span class="ct-series-a"></span>series a</li>`
`<li><span class="ct-series-b"></span>series b</li>`
`<li><span class="ct-series-c"></span>series c</li>`
`</ul>`

gionkunz:
I'm still not sure if this is something one should just do in his project by using Sass, or if we should include something that is readily available using CSS. It's putting some potentially dead code onto us without really having a Legende feature directly in Chartist. Maybe that's something the legends plugin should worry about by extending the Sass file of Chartist? https://github.com/CodeYellowBV/chartist-plugin-legend

gionkunz:
I think this should go into the legend plugin using Sass settings

PR #593, opened 2016-01-22
Implement stacked line charts

morleyzhi:
I noticed that Chartist allows stacked bar charts, but not line charts. So I added this feature for line charts, along with an example and test.

The implementation is a little blunt: during the render step, I modify the data object directly before all the other actions happen. I need to do this to get the axes do adapt to the data changes. I hope this is okay.

As always, feedback is appreciated!

:thumbsup:

### DIFF
--- a/site/data/pages/examples.yml
+++ b/site/data/pages/examples.yml
@@ -11,6 +11,13 @@ sections:
           intro: An example of a simple line chart with three series. You can edit this example in realtime.
       - type: live-example
         data:
+          title: Stacked line chart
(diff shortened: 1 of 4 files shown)


morleyzhi:
In using this in my use case, I think the accumulation might not be working properly. I'm going to retract this and make sure it's right.

morleyzhi:
I realized that it's better if the developer implements line-stacking in their data, so this is moot. Wish I had realized that earlier, but oh well. Thanks!

PR #596, opened 2016-01-23
Fix for 464: include zero in ct:value

### LINKED ISSUE #464: ct:value Not Set When Value=0
On a line chart, it appears as though the `ct:value` is not set if the value is 0. Here's an example line where I use object notation for setting the value and meta of the object:

``` javascript
overview_data.series[0].push({meta : i, value: campaign.y})
```

If I run this and `campaign.y` is 0, then it results in the following:

image

This works just fine when `value` != 0. This typically wouldn't be an issue, but tooltips tend to try and grab the value to display and, if it's 0, it returns undefined. 

Not sure if you'd consider this a bug or not, just wanted to bring it to your attention.

Keep up the great work!

psalaets:
Changes
- `ct:value` attribute of `line.ct-point` in Line charts no longer omits 0 values
- `ct:value` attribute of `line.ct-bar` in Bar charts no longer omits 0 values

I'm not completely sure if this change is appropriate for Bar chart but it seemed like it. I can remove that from PR if it should not be in there.

Fixes #464

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -346,9 +346,7 @@
 
         // Create bar element
         bar = seriesElement.elem('line', positions, options.classNames.bar).attr({
-          'value': [value.x, value.y].filter(function(v) {
-            return v || v === 0;
-          }).join(','),
+          'value': [value.x, value.y].filter(Chartist.isNum).join(','),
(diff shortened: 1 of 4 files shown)


gionkunz:
Awesome piece of work! Cheers Paul!

PR #615, opened 2016-02-17
Replace ct-point element. Use svg circle instead of line.

AlexanderKozhevin:
1. This minor change will allow to use cool css animation for the **ct-point**. 
2. Ct-point is a point, why we should use line instead?)

I replaced element and fixed proper styles only. 

hover


gionkunz:
There is no way to specify the radius in CSS if you're using circles. When using lines with 0 length you can control the size using stroke-width. You can also control in css then if you'd like to use circle shapes or square shapes for your points by using the stroke-linecap CSS property.

Using lines for dots puts you into the nice situation that you have much to customize using CSS which is preferred within the Chartist project. We'll therefore not switch to circles for points.

AlexanderKozhevin:
Nope, actually we can specify "r" property in css. 
Here is a [fiddle](https://jsfiddle.net/AlexanderKozhevin/aapycL87/).

gionkunz:
Yes, but only in SVG 2 which we can't rely upon at the moment. Your fiddle is working in Chrome but not in Firefox nor IE at the moment. SVG 1.1 does not specify `r` as presentational attribute that can be styled with CSS https://www.w3.org/TR/SVG/styling.html#SVGStylingProperties

AlexanderKozhevin:
Ops, I didn't check it. Thanx for note.

PR #623, opened 2016-02-25
Fixed bug with negative start angle getting set to 0 for donut charts

cw882:
Bug introduced in commit ec00b726030a8b5ec90b263a0cca7a65dc98b4f3

e.g. If I pass a start angle of -150, this is getting set to 0 by the Math.max() function.

### DIFF
--- a/dist/chartist.js
+++ b/dist/chartist.js
@@ -3971,7 +3971,7 @@ var Chartist = {
       var endAngle = startAngle + dataArray[i] / totalDataSum * 360;
 
       // Use slight offset so there are no transparent hairline issues
-      var overlappigStartAngle = Math.max(0, startAngle - (i === 0 || hasSingleValInSeries ? 0 : 0.2));
+      var overlappigStartAngle = Math.max(-360, startAngle - (i === 0 || hasSingleValInSeries ? 0 : 0.2));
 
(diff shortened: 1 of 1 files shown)


gionkunz:
CC @hansmaad 
Is this still an issue with the current version?

hansmaad:
@gionkunz I think so. But I'm not sure if this fix is enough to allow negative angles. I could check this out if we want to support it. And we have to add some tests...

gionkunz:
I'd really appreciate a fix for this. I think negative angles should be supported.

hansmaad:
I don't know why this doesn't happen on the CI build, but the change breaks 2 tests on my local build.

x Pie chart tests Pie with small slices Pie should render correctly with very small slices
x Pie chart tests Pie with small slices Donut should render correctly with very small slices

Screenshots from Chrome:
After change:
hairline

Before change:
hairline_before

The `overlappigStartAngle` was reset to 0 before. It's now < 0 and thus creates this visible hairline.
I think there has to be a check that `overlappigStartAngle` is not less than the previous `startAngle`.

gionkunz:
@cw882 Can you do more tests and check with the suggestion of @hansmaad ?

gionkunz:
@cw882 any progress on this?

Issue #626, opened 2016-02-29
Monotone cubic interpolation


ammojamo:
This is related to line smoothing issues already discussed at #92 and #155.

I've just implemented Monotone Cubic Interpolation for Chartist, using the Fritsch-Carlson method, based on this math.stachexchange post: http://math.stackexchange.com/questions/45218/implementation-of-monotone-cubic-interpolation

The main problem I was trying to solve was where the extrema of the fitted curve lie beyond the extrema of the original data points. This is particularly bad when the curve dips below zero for a quantity that is inherently non-negative, e.g. number of units sold.

Results are pretty promising - unlike 'simple' interpolation it doesn't put artificial kinks into the curve, whilst still preserving monotonicity of the original data points.

<img width="1312" alt="monotonecubic" src="https://cloud.githubusercontent.com/assets/89878/13383119/6967f4aa-ded8-11e5-8bb6-963ef26f7e78.png">

Would you like me to submit a PR?

gionkunz:
Hey there. Thanks for your contribution! Yes, sure. I would love to see that PR :-)

farnabaz:
@ammojamo @gionkunz  It will be great if you rebuild in order to update library, by now this PR only applied in source of library

gionkunz:
Hi sorry thats my fault. Ill publish a release the upcoming week.

PR #627, opened 2016-02-29
Monotone cubic interpolation

ammojamo:
See: Issue #626

The code is modelled after the existing cardinal interpolation code. For example I have used a similar splitIntoSegments() function, but with a small modification.

Actual interpolation logic is pretty straightforward to follow.

Not sure about updating tests or documentation, let me know if anything's needed.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -995,4 +995,71 @@ var Chartist = {
     };
   };
 
+
+  /**
+   * Splits a list of coordinates and associated values into segments. Each returned segment contains a pathCoordinates
+   * valueData property describing the segment.
+   *
+   * With the default options, segments consist of contiguous sets of points that do not have an undefined value. Any
+   * points with undefined values are discarded.
+   *
+   * **Options**
+   * The following options are used to determine how segments are formed
+   * ```javascript
+   * var options = {
+   *   // If fillHoles is true, undefined values are simply discarded without creating a new segment. Assuming other options are default, this returns single segment.
+   *   fillHoles: false,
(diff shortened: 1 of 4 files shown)


MightyPork:
Hoping this gets merged soon, it's really good!

I've patched my chartist and using this in my project, much better than the other methods.

ammojamo:
@gionkunz Is there any issue with this PR preventing it from being merged? Anything I can do to help?

gionkunz:
Hi there thanks for the PR! This is great! My only concerns are possible code duplication and missing tests. Maybe its time to find a general `splitIntoSegments` that goes into `Chartist.core` and can be used by cardinal as well as monotone cubic interpolation. Also tests should be written for the line chart spec. We already have a few tests that include interpolation.

ammojamo:
Cool, I'll have a go at fixing these issues

ammojamo:
I've now refactored `splitIntoSegments` into core (with docs and tests) and also added a test for the interpolation which uncovered a small bug, also fixed.

I noticed that the API documentation for `splitIntoSegments` doesn't get generated correctly, but this is a actually a wider issue with `dox` which is already affecting your API docs. I've submitted a PR with dox for this (https://github.com/tj/dox/pull/177)

ammojamo:
@gionkunz Just checking if you've had a chance to review the latest changes on this PR? Let me know if anything else needs attention.

gionkunz:
Awesome man! This looks perfect!

gionkunz:
I believe we can get that straight in. Thanks for your work!

PR #629, opened 2016-03-01
Prevent negative value for foreignObject width attribute

joserobleda:
When I create a line chart with a custom offset like this:

``` js
Chartist.Line(Node, {labels, series}, {
  axisY: {
    offset: 0
  }
})
```

I'm getting this error:

image

I can't figure it out why this `- 10` [in this line](https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/core.js#L885) is necessary, but looks like it is some hard coded value, and removing it resolves problem.

I'm presume this will cause a visual change for the apps which are currently using the lib. Anyway, I've readed the contributing file, but let me know if this PR is ok

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -882,7 +882,7 @@ var Chartist = {
     positionalData[axis.units.pos] = position + labelOffset[axis.units.pos];
     positionalData[axis.counterUnits.pos] = labelOffset[axis.counterUnits.pos];
     positionalData[axis.units.len] = length;
-    positionalData[axis.counterUnits.len] = axisOffset - 10;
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi thanks for the PR! The hardcoded `-10` does optically correct the label position. You are right with your assumption that removing this will invluence the visual appearance of any existing chart. We should probable better go with `Math.max(0, axisOffset - 10)`

joserobleda:
:+1:

joserobleda:
> We should probable better go with Math.max(0, axisOffset - 10)

Done!

gionkunz:
Cool, thanks!

rodrigm:
:heart_eyes:

PR #636, opened 2016-03-11
Fix updateCurrentOptions to expect a media query event

### LINKED ISSUE #525: optionsChanged event never emitted
The 'optionsChanged' event is never emitted when a chart changes due to media query config being applied. This is because in Chartist.optionsProvider, the function updateCurrentOptions expects to be passed an optional boolean, but in fact when called from a media query listener, it is passed an event object, therefore tripping over the event emit guard.

I have the following fix:

https://github.com/BlackPepperSoftware/chartist-js/commit/2e3685ca06b6aa681c559f74841d627902d59184

But I couldn't get the project to build owing to libsass / node-sass errors, so I haven't submitted a pull request.

stringfellow:
fixes #525

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -886,7 +886,7 @@ var Chartist = {
       mediaQueryListeners = [],
       i;
 
-    function updateCurrentOptions(preventChangedEvent) {
+    function updateCurrentOptions(mediaEvent) {
       var previousOptions = currentOptions;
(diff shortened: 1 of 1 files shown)


gionkunz:
Great! Thanks!

stringfellow:
@gionkunz thanks so much for merging this - do you have any idea when you might include it in a minor release? I've tried building it but am having troubles and would love to be able to make use of this event being emitted!

PR #637, opened 2016-03-14
Zoom plugin 0.2.1

### LINKED ISSUE #535: Zoom plugin example does not work on Firefox
The zoom plugin does not work on the plugins example page on Firefox (42.0). Since the plugin works on Firefox in general, I would track this here. There is a `TypeError: e is null` in mouse event handlers. Works in Chrome 47, IE11 (all Win7 64bit).

hansmaad:
Fixes gionkunz/chartist-js#535.
Update zoom plugin example to use 0.2.1 which has several fixes. Example now works in Firefox.


gionkunz:
Cool!

PR #645, opened 2016-03-18
Fix #643 Prevent infinite loop in getBounds if bounds.valueRange is very small

### LINKED ISSUE #643: Infinite loop in getBounds if bounds.valueRange is very small
Can be reproduced on example page: http://gionkunz.github.io/chartist-js/examples.html
Edit the "SIMPLE LINE CHART" example and replace series with:

```
series: [ [1, 1.0000000000000002] ]
```

Open Chrome task manager and kill the process :-)

In this case `bounds.step` which is calculated from the very low order of magnitute is less than `Number.EPSILON` and `newMin + bounds.step` is always `newMin`.
This loop runs for ever: 

```
while(newMin + bounds.step <= bounds.low) {
   newMin += bounds.step;
}
```

https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/core.js#L733

A quick fix is to add `bounds.step = Math.max(bounds.step, Number.EPSILON);` right above this loop, but I'm not sure if this is the best way to handle such cases.

hansmaad:
Fixes #643.
This fix added a simple check, if `bounds.step` is less than EPSILON.
Additionally if filters duplicate values in `bounds.values`.

I added some unit test, to ensure the current behaviour, even if I'm not sure if it's always the desired behaviour (Are there any other open issues here?).

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -727,6 +727,10 @@ var Chartist = {
       }
     }
 
+    // step must not be less than EPSILON to create values that can be represented as floating number.
+    var EPSILON = 2.221E-16;
+    bounds.step = Math.max(bounds.step, EPSILON);
+    
     // Narrow min and max based on new step
     newMin = bounds.min;
     newMax = bounds.max;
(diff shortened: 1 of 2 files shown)


gionkunz:
Awesome! Thanks!

src/scripts/core.js
```diff
@@ -727,6 +727,10 @@ var Chartist = {
       }
     }
 
+    // step must not be less than EPSILON to create values that can be represented as floating number.
+    var EPSILON = 2.221E-16;
```
gionkunz, line 731: Why are we not using Number.EPSILON ?

gionkunz, line 731: Ah sorry this is ES6. Ignore my comment :-)

Issue #650, opened 2016-03-27
Proposed fix for blurred grid lines


MightyPork:
Sorry for not making a PR, but I couldn't get grunt working so couldn't test it anyway - and honestly, I'm too lazy. You'll patch it quickly manually:

``` js
  Chartist.createGrid = function(position, index, axis, offset, length, group, classes, eventEmitter) {
    var positionalData = {};
    positionalData[axis.units.pos + '1'] = Math.round(position)+0.5; // EDITED: fix blurred grid
    positionalData[axis.units.pos + '2'] = Math.round(position)+0.5; // EDITED: fix blurred grid
```

(this "patch" is against the latest release dist)

If you round the value and add 0.5, the grid won't be blurred in Chrome. 
It's blurred in all the examples, but even worse looking if you have it white on black.

Compare:

![screenshot_2016-03-27_21-13-28](https://cloud.githubusercontent.com/assets/2041118/14067244/c96bcff0-f460-11e5-81ee-1b1648e87238.png)

![screenshot_2016-03-27_21-13-13](https://cloud.githubusercontent.com/assets/2041118/14067243/c72dbdc0-f460-11e5-9782-dad584124dd4.png)

gionkunz:
We've decided to go with `shape-rendering: crispedges;` since this is up to the rendering engine how to handle it. What happens if the grid line is 2px or any other even number? Using the CSS approach, we can fix it in the rendering engine.

PR #654, opened 2016-03-29
Avoid blurred grid lines

hansmaad:
For issue https://github.com/gionkunz/chartist-js/issues/650
Looks good on Chrome. Doesn't seem to help for Firefox and IE. Not testes on Safari...

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -747,7 +747,9 @@ var Chartist = {
     var values = [];
     for (i = bounds.min; i <= bounds.max; i += bounds.step) {      
       var value = Chartist.roundWithPrecision(i);      
-      value != values[values.length - 1] && values.push(i);
+      if (value !== values[values.length - 1]) {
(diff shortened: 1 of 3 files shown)


gionkunz:
Hey thanks for the PR. You checked in an fdescribe. Im not sure about this fix though as rendering behavious is different on different browsers. I think the preferred way is to use CSS: `shape-rendering: crispEdges` on grid lines and rather stick to the exact calculated values.

hansmaad:
I'm not sure if `shape-rendering: crispEdges` always helps. My best practice when drawing vertical or horizontal svg lines is to round to .5 as proposed in #650. But you're right, that the behaviour depends on the browser. 
If you feel better to keep the old implementation I would make another commit, so that you can merge the tests I added.

gionkunz:
I'm closing this since I really believe we should paint accurate and let the browser deal with the subpixel stuff.

Issue #634, opened 2016-03-07
Charts have zero width/height in Pale Moon


David-Hari:
![Zero size](http://s24.postimg.org/cxqc75uzp/zero_size.png)

Was working fine up until a few weeks ago.

I reported it at the [PM forums](https://forum.palemoon.org/viewtopic.php?p=79561#p79550), thinking it was likely a problem with their rendering engine. But it might be worth you having a look in case there is a work-around, or perhaps your current implementation isn't as robust as it could be.

I remember seeing another issue here where you mentioned a width/height hack for an older version of Firefox. Could be a related issue (as Pale Moon is related to Firefox).

derekblank:
@David-Hari: I noticed this identical issue when testing older versions of Firefox (dating back to at least version 32.0.2) with the latest version of Chartist (0.9.7). Upgrading to the latest version of Firefox (45.0.1 ) resolved the issue immediately, which might suggest that this is an issue with PaleMoon rendering.

David-Hari:
Ok, good to know. Hopefully Pale Moon will fix the issue soon.

gionkunz:
Closing becauase of inactivity

Issue #592, opened 2016-01-22
Width and height not working in Firefox


albertolopez:
When setting width and/or height in the options of any chart, the chart is cropped but not resized.

I am using Firefox 43.0.4

![chartist - examples 2016-01-22 10-26-39](https://cloud.githubusercontent.com/assets/309681/12506924/ac7ef864-c0f2-11e5-87a0-a31a46f36f4e.png)

The code is the one in the examples, I have tried with different chart types and all fail

``` javascript
var data = {
  series: [5, 3, 4]
};

var sum = function(a, b) { return a + b };

new Chartist.Pie('.ct-chart', data, {
  width: 300,
  height: 300,
  labelInterpolationFnc: function(value) {
    return Math.round(value / data.series.reduce(sum) * 100) + '%';
  }
});
```

gionkunz:
Hi there. The problem boils down to the width and height estimations within the Svg DOM builder of Chartist which is rather complex. We included a workaround for getting around this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=530985

However, the bug is long fixed and we should be fine by just relying on `getBoundingClientRect`. I think IE9 should also be fine with it.

I guess https://github.com/gionkunz/chartist-js/blob/develop/src/scripts/svg.js#L338-L351 can be simplified by just using `this._node.getBoundingClientRect().width` and `height`. Also the fallback to `getBBox()` can be removed I guess.

I'll flag this as a bug but happy if someone could pick this up and make a PR. I guess testing in browsers >1%, -IE8, +IE9 would need to be done extensively.

gionkunz:
For now, as the fallback that will be used in Firefox is the width and height of the parent element, just make sure you also have a parent element (the container of the SVG) that holds the desired dimensions.

gionkunz:
I've tested IE9, IE10, Safari 7.1, Firefox, Chrome and Android and going with `getBoundingClientRect` without any fallback does the job. I'll make a quick fix for this.

NoBodyButMe:
Hello

I'm sorry to ask, but is this bug already fixed in the release? I tried to use it but still have the same problem in Firefox.

PR #658, opened 2016-03-30
Jolla browser compatibility

cheese83:
getBoundingClientRect returns 0 height and width on the [Jolla](https://jolla.com/) [browser](https://github.com/sailfishos/sailfish-browser), which causes charts that don't have dimensions set in the options to end up drawn entirely in the top left corner. Presumably this is due to [Mozilla bug 530985](https://bugzilla.mozilla.org/show_bug.cgi?id=530985). It can be worked around by using `window.getComputedStyle` to get the dimensions of the svg element.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -786,8 +786,8 @@ var Chartist = {
     var yAxisOffset = hasAxis ? options.axisY.offset : 0;
     var xAxisOffset = hasAxis ? options.axisX.offset : 0;
     // If width or height results in invalid value (including 0) we fallback to the unitless settings or even 0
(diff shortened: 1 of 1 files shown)


cheese83:
I expect this will also fix #634

gionkunz:
Check this issue: https://github.com/gionkunz/chartist-js/issues/592

gionkunz:
We have removed old browser support and to be honest I'd rather simplify our code than support old browsers. The question is why Jolla uses an old version of the Gecko engine. If we start to support old browsers, we kind of violate our goal to be a library for the modern web.

cheese83:
Ok, that's understandable.

PR #669, opened 2016-04-13
Chartist.extend performance

cheese83:
While writing a page that draws some charts with over 1k points each, I noticed that Chartist.extend was taking up a lot of time, so I optimized it.

I have attached a simple test page that logs the time to draw one chart with 2k points. On my machine I get results like this (milliseconds, averaged over three trials):

```
            Original    Optimized
Firefox 45       229          140
Chrome 49        122           91
IE 11            380          166
```

[test.zip](https://github.com/gionkunz/chartist-js/files/217762/test.zip)

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -56,18 +56,20 @@ var Chartist = {
    * @return {Object} An object that has the same reference as target but is extended and merged with the properties of source
    */
   Chartist.extend = function (target) {
+    var i, source, sourceProp;
     target = target || {};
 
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there. Sorry of the ultra long delay on this :-) I usually tend to prefer functional over imperative style just for readability and maintenance. But since this is critical function in the core, I agree it should be written in a performance sensitive way, and get rid of any unnecessary closure creations.

PR #686, opened 2016-04-29
Prototype.js compatibility

cr3ative:
For some reason, Prototype.js removes the ability for .filter() to use the third argument to refer back to the object it is modifying, used here as "self".

Yes, this is entirely the fault of Prototype.js; this shim simply allows Chartist to operate in this broken environment.

I submit this with the hope, merged or not, it prevents at least one person tearing their hair out when having to find out what the heck was going on.

Huge love for Chartist, by the way. What a wonderful library!

### DIFF
--- a/dist/chartist.js
+++ b/dist/chartist.js
@@ -1993,13 +1993,10 @@ var Chartist = {
    * @return {Chartist.Svg} The wrapper of the current element
    */
   function addClass(names) {
-    this._node.setAttribute('class',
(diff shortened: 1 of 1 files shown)


gionkunz:
Sorry about getting back to you that late and thanks for the kind words :-) I was never a fan of how intrusive Prototype.js presents itself :-) However, if it's only this I believe we can live with a slight change.

PR #719, opened 2016-06-17
Added option to ensure target tick value in AutoScaleAxis

blzzz:
Sometimes you'd like to have a certain number included in the calculated ticks of a AutoScaleAxis. A popular use case for this is the zero axis grid line which is going to disappear if the scale – because of its given data – renders ticks like `[-3,-1,1,3]`. 
To ensure the include of zero the new introduced axis option will help: `ensureTickValue: 0` . Once defined, the values of the bounds are going to be re-set by stepping up and down from the target value (`0` in this case) while the already calculated step size is preserved.

### DIFF
--- a/src/scripts/axes/auto-scale-axis.js
+++ b/src/scripts/axes/auto-scale-axis.js
@@ -14,6 +14,8 @@
  *   onlyInteger: true,
  *   // The reference value can be used to make sure that this value will always be on the chart. This is especially useful on bipolar charts where the bipolar center always needs to be part of the chart.
  *   referenceValue: 5
+ *   // If the tick value is set, its number is going to be included in the axis' ticks
+ *   ensureTickValue: 0
  * };
  * ```
  *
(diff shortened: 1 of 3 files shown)


benib:
@gionkunz what is your general mood about this feature? We have the situation sometimes, that there are charts where the tick at `0` is missing, that makes the chart look pretty weird. So we would love to have this feature in upstream chartist instead of having to maintain a fork.

gionkunz:
Currently I still don't get the requirement right i think. Does `referenceValue` not already cover that sort of requirement?

http://jsbin.com/wajodo/edit?html,js,output#

gionkunz:
maybe a small mockup chart with some annotations would help

gionkunz:
http://jsbin.com/wajodo/edit?html,js,output#

blzzz:
@gionkunz Here's a small mockup chart based on your example: http://jsbin.com/fufevozevi/edit?html,js,output

The chart's data series contains values that result in an y axis scale excluding the zero tick. Obviously one of the reasons for this behaviour has to do with the chart's limited height (see the element's style attribute in the HTML section). By setting `scaleMinSpace` to a lower value than its default (20) this can be fixed, but in our case, we'd like to have a static value there. As you can see, the option `referenceValue` set to zero won't fix this.

gionkunz:
I'd rather appreciate a tick override mechanism on the `AutoScaleAxis`. This would make things much easier and we don't need to bloat the `getBounds` function more than necessary. I've create a branch to test this with the following commit: https://github.com/gionkunz/chartist-js/commit/92e1adca01b4477937949f19e63020e17ea0b7db

The only downside, if this is relevant, would be that if ticks are specified that are out of the bounds, they're not displayed. To change this we could use the ticks to project values similarly like we do for FixedScaleAxis.

gionkunz:
Please feel free to comment, make any modifications and submit a PR that uses the above test code.

blzzz:
Thanks for you suggestion. I just submitted a new PR: https://github.com/gionkunz/chartist-js/pull/736

gionkunz:
Closing this in favor or #736

PR #724, opened 2016-06-23
Fix infinite loop in getBounds with a more robust increment.

hansmaad:
This is related to bfdc8f2 which introduced EPSILON as the smallest increment but fails for bigger numbers.
See https://github.com/gionkunz/chartist-js/pull/645

This fix uses `value *= (1 + EPSILON)` instead of `value += EPSILON` to prevent inifinite loops in `getBounds`. EPSILON is only safe when incrementing 1.

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -727,25 +727,31 @@ var Chartist = {
       }
     }
 
-    // step must not be less than EPSILON to create values that can be represented as floating number.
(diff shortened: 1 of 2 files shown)


gionkunz:
Thanks!

gionkunz:
I've already talked about this in a different issue, but we should really try to think about a simplification of the getBounds function.

PR #727, opened 2016-06-27
Set the summary option to jasmine task to better analyse failed tests.

hansmaad:
This option displays a list of all failed tests and their failure messages.
We won't have to scroll the output for failures any more.


gionkunz:
Makes sense! Cheers

Issue #711, opened 2016-06-07
Grid background


hansmaad:
I wonder what's the preferred way to add a fill to grids. There is no `rect` or similar element to style. I've built a simple example how to add a rect here: http://codepen.io/hansmaad/pen/qNOmNV 

```
chart.on('created', function(data) {      
    var grids = data.svg.querySelector('.ct-grids');
    var chartRect = data.chartRect;
    rect = grids.elem('rect', {
      x: chartRect.x1,
      y: chartRect.y2,
      width: chartRect.width(),
      height: chartRect.height(),
    }, 'ct-fill-rect', true);
});
```

We could provide a little plug-in, but shouldn't this be a build-in feature?

gionkunz:
I guess this should be built in with an option. `showGridBackground` or something similar.

hansmaad:
@gionkunz I started to implement this. A few questions:

Name of option? `showGridBackground` ?
class name for rect? `ct-grid-background` ?
default css? `fill:none` ?
should a draw event be emitted?
Where could I put a `createGridBackground` function to share between `line` and `bar`? In core (like `Chartist.createGrid`)? 

``` javascript
 if (options.showGridBackground) {
      // Draw the chart background
      var gridBackground = gridGroup.elem('rect', {
        x: chartRect.x1,
        y: chartRect.y2,
        width: chartRect.width(),
        height: chartRect.height(),
      }, options.classNames.gridBackground, true);

      // Event for grid background draw
      this.eventEmitter.emit('draw', {
        type: 'grid-background',
        group: gridGroup,
        element: gridBackground
      });
    }
```

gionkunz:
Yes, Yes, Yes, Yes, Yes :-) Awesome stuff @hansmaad 

One small thing, we've used camel case for casing in event types so far: https://github.com/gionkunz/chartist-js/blob/d12e272cf5177280054dddef11634d188dacf4ab/src/scripts/svg.js#L437

dguayrobotiq:
@hansmaad @gionkunz This is kind of old but I'd like to use this but be able to use different color for different bar on my bar graph. Concrete example would be to have a graph bar representing a week with weekend grid background in gray and weekday with white grid background.

PR #735, opened 2016-07-04
Add grid background

hansmaad:
Implements #711 
Adds an option `showGridBackground` to line and bar charts (default is `false`).
If `true`, adds a `.ct-grids rect.ct-grid-background`.


gionkunz:
Fantastic work @hansmaad !

PR #736, opened 2016-07-05
Add option for more tick control to AutoScaleAxis

blzzz:
I think a feature that enables to override calculated ticks with "static ticks" would not really help us in our case (it maybe helps in other cases). In the example I showed in the other PR (http://jsbin.com/fufevozevi/edit?html,js,output) the resulting scale of the y axis does not include zero (`-1,1,3,5,7`), however we don't want it to become something like this instead: `-1,0,1,3,5,7`. With your suggested override feature, this could now easily be achieved by setting `axisY:{ ticks:[0] }`.

We'd rather like to preserve a regular interval to get a scale like this: `-2,0,2,4,6,8` . That means it's not enough to only set/override one (or more) ticks, but the new ticks also need to get recalculated dynamically in stepping up and down from a tick value by the given interval. So it was my concept to not only ensure a certain tick value, but also to step up and down from it. In the end, the calculation results in complete new ticks.

I understand you not wanting to include this in the `getBounds()` function because this could lead to a certain overload. What I suggest you instead is to insert the recalculation of the ticks as part of the AutoScaleAxis where you suggested the tick override feature. I know that the code is still quite similar to what I suggested before, but I still think this is the way to go to solve this issue. Please let me know what you think about it.

### DIFF
--- a/src/scripts/axes/auto-scale-axis.js
+++ b/src/scripts/axes/auto-scale-axis.js
@@ -39,7 +39,7 @@
 
     if (!isNaN(parseFloat(options.ensureTickValue)) && ticks.indexOf(options.ensureTickValue) == -1) {
 
-      // redefine ticks by ensuring a given tick value and by preserving the calculated step size / interval
+      // Redefine ticks by ensuring a given tick value and by preserving the calculated step size / interval
       ticks = [];
       var currTick = options.ensureTickValue;
       var tickStep = this.bounds.step;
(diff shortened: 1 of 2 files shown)


gionkunz:
Looks way cleaner! Thanks. Can you please add a bunch of tests to cover the various use-cases?

blzzz:
Cool! Tomorrow I'll find time to add a bunch of tests.

blzzz:
@gionkunz what do you think about the tests? is it what you expected?

IDisposable:
:shipit:

blzzz:
@gionkunz what's the state of this PR?

Issue #714, opened 2016-06-13
Default SCSS only colors up to "o" in a series (disappearing bars)


afiaka87:
Was having a lot of trouble figuring out why certain bars in my chart weren't showing up. Found out it was because the CSS for a one dimensional series only colors up to the generated "o" series name. 

> // Series names and colors. This can be extended or customized as desired. Just add more series and colors.
> $ct-series-names: (a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) !default;
> $ct-series-colors: (
>   #d70206,
>   #f05b4f,
>   #f4c63d,
>   #d17905,
>   #453d3f,
>   #59922b,
>   #0544d3,
>   #6b0392,
>   #f05b4f,
>   #dda458,
>   #eacf7d,
>   #86797d,
>   #b2c326,
>   #6188e2,
>   #a748ca
> ) !default;

I realize there's a comment specifying that it can be modified, but this caused me a good deal of troubleshooting as my bar chart had values which weren't getting a stroke, hence invisible.  At the very least, the default series should extend to 'z'.

gionkunz:
Hi there, I'm not sure if chartist is made for such large count of series. However, I agree that we should include proper default colours so at least we're getting visible series. Anyway, we're going to refactor the series names to be numeric and with this change probably also re-evaluate how we're generating colours. I'd like to include an option to automatically create series colours based on scales and colour spaces like LAB https://en.wikipedia.org/wiki/Lab_color_space

gionkunz:
Parts of this we're keeping track in #183, I'll close this issue in favor of the series renaming and keep a reference to it.

afiaka87:
Thanks. For what it's worth, after re-styling them to have colors, I think Chartist works really well with a large count of series, especially when making use of vertical space.

Issue #451, opened 2015-09-13
Pie breaks after the 15th color


dreamyguy:
The pie seems to break when I set a big amount of data. After the 15th color the pie goes transparent (donut) or black (default) for a while, and then picks up again showing 15 colors - to break again and so on.

The graphs below show the total number of commits on 346 repos.

![pie-bug-after-15th-color-a](https://cloud.githubusercontent.com/assets/1021849/9838495/226896c0-5a5f-11e5-8e3d-14531ad378ea.png)

![pie-bug-after-15th-color-b](https://cloud.githubusercontent.com/assets/1021849/9838553/ce324860-5a60-11e5-89f7-3fd08c79be2c.png)

This is the setup (lipsum-ized for the sake of privacy):

```
new Chartist.Pie('.ct-chart', {
[...]

Issue #354, opened 2015-06-17
Empty spaces in stacked bar chart


wedgef5:
I'm having a problem with a stacked bar chart, where there are large blank areas in the bars.  When I inspect these elements using Firebug, I can see that they are being rendered as 0 height, however, their y1/y2 values show they they are taking up space in the chart.  It seems that several of these are adding up to form the blank area.

A couple of ideas I have had as to possible causes: (1) I am passing the data for the series in as an array of objects with 'name' and 'data' attributes.  I saw this done explicitly in a Line graph example.  I assumed this was OK for a Bar graph, but I'm not 100% sure, and (2) the data ranges very widely, going from 0 to 10^9.  I wasn't sure is scaling over such a wide range might cause problems.

Here is a screen shot showing the problem.  Any suggestions are appreciated!

![stacked_bar](https://cloud.githubusercontent.com/assets/12939463/8213611/a884c252-1512-11e5-9776-0131f3b63a7b.png)

georgerappel:
Could you show me your graph code, @wedgef5 ?

gionkunz:
I saw negative values causing similar issues. Are you sure you only have positive values in your data?

wedgef5:
Sorry for the slow reply...I've been out of the office for a couple of days.

@georgerappel - here is the 'success' section of my AJAX call...

``` javascript
        success: function(data) {
            chart = new Chartist.Bar('.ct-chart', data, {
                stackBars: true,
                axisY: {
                    labelInterpolationFnc: function(value) {
                        return (value / 1000) + 'k';
                    }
                },
                low: 0
                }).on('draw', function(data) {
                  if(data.type === 'bar') {
                    data.element.attr({
                      style: 'stroke-width: 30px'
                    });
                  }
            });
            var $tooltip = $('<div class="tooltip tooltip-hidden"></div>').appendTo($('.ct-chart'));

            $(document).on('mouseenter', '.ct-bar', function() {
                var seriesName = $(this).closest('.ct-series').attr('ct:series-name'),
                    value = $(this).attr('ct:value');
                $tooltip.html(seriesName + '<br/>' + value);
                $tooltip.removeClass('tooltip-hidden');
            });

            $(document).on('mouseleave', '.ct-bar', function() {
                $tooltip.addClass('tooltip-hidden');
            });

            $(document).on('mousemove', '.ct-bar', function(event) {
                $tooltip.css({
[...]

PR #745, opened 2016-07-21
Add default colors to total of 26, so there won't be unrendered lines

lauri-kaariainen:
This works as a non-breaking (albeit might change colours for some) fix for the problem that ct-series from -p to -z are left blank because of missing stroke colors.
Fixes permanently problems like with:
- #354 
- #451 
- #714 

I also spent good many hours identifying this problem, it was hard to spot because it became apparent only after 15 series and was very unexpected of charting library.

I took the colors from here:
[http://jsbin.com/zozelilomi/edit?css,js,output](http://jsbin.com/zozelilomi/edit?css,js,output)


IDisposable:
:shipit:

Issue #744, opened 2016-07-20
Grunt failing when installing locally


drewatk:
Hello! I'm trying to install the dev dependencies in order to contribute, but grunt is failing when I run `grunt preview` I tried `--force` but that gave even more errors. Are the contributing docs out of date? Am I missing something?

Thank you!

```
❯ grunt preview                                                                     chartist-js/git/develop !
Running "clean:public" (clean) task
>> 0 paths cleaned.

jit-grunt: Plugin for the "assemble" task not found.
If you have installed the plugin already, please setting the static mapping.
See https://github.com/shootaroo/jit-grunt#static-mappings

Warning: Task "assemble" failed. Use --force to continue.

Aborted due to warnings.


Execution Time (2016-07-20 19:10:00 UTC-4)
loading tasks                78ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 82%
loading grunt-contrib-clean   3ms  ▇▇ 3%
clean:public                 14ms  ▇▇▇▇▇▇▇▇▇▇ 15%
Total 95ms
```

drewatk:
I've added the `grunt-assemble` plugin to devDependencies as pull request #746

PR #746, opened 2016-07-24
Adds grunt-assemble plugin to devDependencies

drewatk:
Adding this dependency to the package.json allows the grunt task "assemble" to succeed. Currently (following the contribution guide), It fails when running `grunt preview` and other commands.

Issue #744


gionkunz:
Great, thanks! I must have missed the --save-dev flag on the recent stack update :-(

Issue #733, opened 2016-07-01
Include most wanted examples on example page


gionkunz:
We should collect a set of examples that are most wanted and then extend our examples:

Here's one:

Time series using fixed scale axis and momentjs: http://jsbin.com/lokozu/edit?html,css,js,output

drewatk:
Hey, I'd be happy to help with this one, would this be above the current line chart, bar chart, etc. examples?

gionkunz:
Hi there. Better would be to add examples to the examples page. Check out data/examples.yml. Cheers

drewatk:
Would the time series go under Line chart examples?

gionkunz:
Yep that's where we should place it for now.

drewatk:
I've opened a pull request for that example

gionkunz:
Using images in labels: http://jsbin.com/dilojuh/1/edit?js,output

gionkunz:
Centered label in donut chart: http://jsbin.com/buhoza/2/edit?css,js,output

gionkunz:
Duplicate of #255

gionkunz:
Adding images to series meta data and use Chartist.Svg and draw event to replace line chart points with images: http://jsbin.com/wadoteg/edit?html,css,js,output

gionkunz:
This bar chart would also be interesting: http://jsbin.com/hoqonab/edit?html,js,output

PR #753, opened 2016-08-06
Add Moment.js Timeseries Example to examples page

drewatk:
I added a commonly requested example, to help with #733. The example code was adapted and simplified from the [jsbin on that issue.](http://jsbin.com/lokozu/edit?html,css,js,output)


gionkunz:
Great thanks!

PR #769, opened 2016-08-22
Fixed a bug of updating chart under scale transforming

Dafrok:
There is a bug when refreshing the chart under scale transforming: the
zoom size of the chart will be square of the scale value.

Demo: http://codepen.io/Dafrok/pen/oLRxka

### DIFF
--- a/src/scripts/core.js
+++ b/src/scripts/core.js
@@ -796,8 +796,8 @@ var Chartist = {
     var yAxisOffset = hasAxis ? options.axisY.offset : 0;
     var xAxisOffset = hasAxis ? options.axisX.offset : 0;
     // If width or height results in invalid value (including 0) we fallback to the unitless settings or even 0
-    var width = svg.width() || Chartist.quantity(options.width).value || 0;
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there. First I really appreciate your efforts and the PR. However, since this is a critical part of the code in terms of cross browser compatibility, we need to be extremely careful about changing it. 

Whats the motivation of this change? I understand the behaviour is kind of weird but what's the use case?

Dafrok:
OK, I'll show my case.

I was developing a full-page-slides-editor application. People can make their own website with several pages by this app. There is a page list in the editor which shows thumbnails of every pages. When users changes or saves these pages, the page list will refresh immediately. Unlucky if there is a charts in pages, the size of thumbnails goes wrong. 

However this PR only works when the 'width' & 'height' options were existing. Maybe there are some better solutions, I'll thinking about it recently, thank u for your work.

LucienLee:
I met the same problem! When chart initiate or update under scale transforming, the size of path would be wrong!

W4G1:
Any updates on this? Having the same problem with scale() as well.

PR #772, opened 2016-08-22
Fix #403 - Add amdModuleId

### LINKED ISSUE #403: Import not working with webpack+babel
I'd like to use chartist in the manner below. My workflow uses webpack and babel.

> import Chartist from 'chartist';

This throws the following in the browser: `Uncaught TypeError: Cannot set property 'Chartist' of undefined`

I have got around this by forking chartist and tweaking the factory code in the manner described here: https://github.com/gregjacobs/Autolinker.js/issues/86#issuecomment-74487812

It would be great if this could be incorporated into the master branch.

chiefcll:
Needed for webpacking scripts - sets name correctly


gionkunz:
Nice! Thanks!

PR #789, opened 2016-09-15
Allow bar charts to have auto scaling axis by overriding referenceValue

sarendipitee:
When you have bar charts with large numbers, the chart becomes pretty useless unless you specifically define a `low` value. 
Example:

screen shot 2016-09-15 at 4 40 15 pm

This simplifies the bar chart `highLow` calc code a bit by adding `referenceValue: 0` as a default, so you don't need to extend the options _again_, and most importantly this allows you to override `referenceValue` to null to get auto scaling back so your chart can be nice without having to explicitly set a low or calculate it outside of chartist.

screen shot 2016-09-15 at 4 50 01 pm

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -65,6 +65,8 @@
     high: undefined,
     // Overriding the natural low of the chart allows you to zoom in or limit the charts lowest displayed value
     low: undefined,
+    // Unless low/high are explicitly set, bar chart will be centered at zero by default. Set referenceValue to null to auto scale.
+    referenceValue: 0,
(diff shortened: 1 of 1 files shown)


sarendipitee:
bump.

gionkunz (changes requested):
Sorry, I was just refactoring the core with a much cleaner way how we process data. Can you please rebase your PR with the latest from develop?

sarendipitee:
Fixed the conflict.

gionkunz (approved):

gionkunz:
Awesome, thanks! Your contribution is much appreciated

Issue #783, opened 2016-09-02
NaN for Piechart with [0,0] as series-data


skuhnow:
I have a dynamically generated Piechart with a series with two values. In the case of both values equals 0 I expect and get an empty chart, but I also get some errors in my console:

`
new Chartist.Pie('.ct-chart', {
    series: [0,0]
}, {});
`

> Error: <path> attribute d: Expected number, "MNaN,NaNA395,395,…".
> Error: <text> attribute dx: Expected length, "NaN".
> Error: <text> attribute dy: Expected length, "NaN".
> Error: <path> attribute d: Expected number, "MNaN,NaNA395,395,…".
> Error: <text> attribute dx: Expected length, "NaN".
> Error: <text> attribute dy: Expected length, "NaN".

gionkunz:
I guess we should check whether all values are 0 and then remove the values when creating the chart. Creating a chart with an empty values array, does not cause any errors. For the moment I suggest you're passing an empty array if all values are 0.

alexandervaneck:
I've addressed this in #797.

PR #797, opened 2016-09-29
Fix for #783 NaN for PieChart with sum values zero

alexandervaneck:
Hi, I was not quite sure how to write a test for this. Since testing console output seems like a bad idea. 

If you have any comments on this implementation I'd love to hear them! 😄 

This fixes and issue presented by #783 where console errors are printed when the total sum of series is 0. Chartist tries to generate an svg to an endpoint NaN, by dividing by 0.

### DIFF
--- a/src/scripts/charts/pie.js
+++ b/src/scripts/charts/pie.js
@@ -163,7 +163,8 @@
         (series.className || options.classNames.series + '-' + Chartist.alphaNumerate(i))
       ].join(' '));
 
-      var endAngle = startAngle + dataArray[i] / totalDataSum * 360;
+      // If the whole dataset is 0 endAngle should be zero. Can't divide by 0.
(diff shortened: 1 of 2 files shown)


gionkunz:
Thanks for the PR! The code looks fine to me, I agree on the use of ternary operator in simple use cases. Can you please add tests to this file? https://github.com/gionkunz/chartist-js/blob/develop/test/spec/spec-pie-chart.js

You can check other tests on how to write them. We should cover at least the use case outlined in the issue #783

gionkunz (changes requested):
Create tests which test for the use-case behind #783

alexandervaneck:
Updated PR, created a test for the usecase described in #783

gionkunz (approved):

Issue #713, opened 2016-06-11
axisY labels are off by one with position 'right' option


ajdust:
It appears that using axisY position 'right' causes the y axis labels to shift up by one incorrectly.

See [here](http://jsbin.com/wiqufijifu/edit?html,js,output) for an example. The 0 label should be on the bottom line.

gionkunz:
I can confirm this bug. Let's fix it ASAP. Thanks for the report!

drewatk:
I looked into this a bit. The axis is only checking for `'start'` and `'end'`, as opposed to `'right'` or `'left'`

An if statement (`axis.js` lines 71 and 92) checks and moves the label offset accordingly either for `'start'` or not start. But a non valid input like `'right'` doesn't match up with the `classNames` object (`bar.js` lines 103-104, `line.js` lines 107-108), so the correct class isn't applied.

@gionkunz What would the best way to approach this be? Throw an error for bad input? Or also accept `'left'` and `'right'` as valid?

gionkunz:
Should be tracked with #811

ariellevitsky:
Sorry, so how would one fix the offset? I checked out the linked issue, but I'm not sure how to apply it to my instance.

PR #799, opened 2016-09-29
Suggested fix for #713 position of axisY labels

alexandervaneck:
If the position is not specifically set to 'start' default the position to 'end'. Any other value is not allowed, both possible values are explicitly mentioned in the documentation.


gionkunz:
Thanks!

gionkunz:
Fixed #713

Issue #716, opened 2016-06-15
Line chart with a single value


erasethis:
If I only have a single value to display the corresponding dot in the chart is drawn to the left of the y-axis, and in the browser console I get messages such as

`chartist.js:1810 Error: <line> attribute x1: Expected length, "NaN".`

gionkunz:
I've confirmed this as a bug. Let's fix this ASAP. Thanks

gionkunz:
Should be fixed once 0.9.9 is released

Webbrother:
Reproduced in 0.10.1, but not sure about reason of such errors.

PR #800, opened 2016-10-04
Fix error in function StepAxis returning NaN 

jmlavoier:
screen shot 2016-10-04 at 5 49 45 pm

In line chart with **fullWidth: true**, the function StepAxis is occurring a division by zero, therefore is returning _NaN_ to attributes **x1** and **x2** of axis.

This PR correct this follow issue: https://github.com/gionkunz/chartist-js/issues/716

### DIFF
--- a/src/scripts/axes/step-axis.js
+++ b/src/scripts/axes/step-axis.js
@@ -24,7 +24,8 @@
       options.ticks,
       options);
 
-    this.stepLength = this.axisLength / (options.ticks.length - (options.stretch ? 1 : 0));
+    var calc = (options.ticks.length - (options.stretch ? 1 : 0));
+    this.stepLength = this.axisLength / (calc === 0 ? 1 : calc);
   }
 
   function projectValue(value, index) {
--- a/test/spec/spec-line-chart.js
+++ b/test/spec/spec-line-chart.js
@@ -502,4 +502,24 @@ describe('Line chart tests', function () {
       });
     });
   });
+  
+  describe('x1 and x2 attribute', function () {
+    it('should contain just a datapoint', function (done) {
+      jasmine.getFixtures().set('<div class="ct-chart ct-golden-section"></div>');
+  
+      var chart = new Chartist.Line('.ct-chart', {
+        series: [[
+          {x: 1, y: 2}
+        ]]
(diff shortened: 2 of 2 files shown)


jmlavoier:
Hi @gionkunz, I hope my PR is fine now. Can you review? Thanks

gionkunz (approved):
Looks perfect now! Thanks for the PR

src/scripts/axes/step-axis.js
```diff
@@ -24,7 +24,8 @@
       options.ticks,
       options);
 
-    this.stepLength = this.axisLength / (options.ticks.length - (options.stretch ? 1 : 0));
+    var calc = (options.ticks.length - (options.stretch ? 1 : 0));
+    this.stepLength = this.axisLength / (calc === 0 ? 1 : calc);
```
gionkunz, line 28: @jmlavoier One thing I noticed, and probably we don't have a test for this yet. But what happens if options.ticks.length is 0 and options.stretch is true?

gionkunz: what about:

``` javascript
  var calc = Math.max(options.ticks.length - (options.stretch ? 1 : 0), 1);
  this.stepLength = this.axisLength / calc;
```

jmlavoier: Well noted @gionkunz, really exists a problem. the calc variable will be -1. 

I think the better solution would be:

``` js
    var ticksLength = options.ticks.length;
    if (ticksLength && options.stretch) {
        ticksLength = ticksLength - 1;
    }
    this.stepLength = this.axisLength / ticksLength;
```

I think I saw this another solution, in another PR sometime ago, but I don't remember who did. By chance would you @AlexvEck?

joaolavoier: I'll create another PR with the fix and tests to this method.

gionkunz:
Thanks! Ill work on something else then

src/scripts/axes/step-axis.js
gionkunz: @jmlavoier @joaolavoier @AlexvEck I've fixed this with 1b32910

Issue #633, opened 2016-03-06
Allow fillHoles option on interpolations when multi values are used


gionkunz:
Currently when multi values are used `{x: 100, y: undefined}` the `fillHoles` option does not work

ammojamo:
This bug is relevant to the `splitIntoSegments` function.

Currently the behaviour checks for `value === undefined` to detect a hole in the data. (see https://github.com/gionkunz/chartist-js/blob/796992d0f83a54c7d38c739bfe2499a08215cf01/src/scripts/interpolation.js#L174)

If we changed this condition to something like `value === undefined || (value.hasOwnProperty('y') && value.y === undefined)`, I think this might solve the problem.

Not sure if this is the most elegant solution or not, it might be duplication of logic found at https://github.com/gionkunz/chartist-js/blob/796992d0f83a54c7d38c739bfe2499a08215cf01/src/scripts/core.js#L425

gionkunz:
We should try to switch all mutli value handling to use https://github.com/gionkunz/chartist-js/blob/796992d0f83a54c7d38c739bfe2499a08215cf01/src/scripts/core.js#L619-L627

ammojamo:
Cool, I was looking for something like that.

Only issue is that it looks like `Chartist.getMultiValue` currently returns `0` instead of `undefined` for all non-numeric falsey values. (contrary to the method documentation). This makes it impossible to detect actual undefined values using that function.

Perhaps `getMultiValue` could take a third parameter `defaultValue` which is returned for non-numeric falsey values. This would conveniently be `undefined` if not specified, or callers could pass `0` to get the current behaviour.

gionkunz:
Yep that should be fine. `defaultValue` in `getMultiValue` would be nice.

gionkunz:
Any help on this would be much appreciated

PR #816, opened 2016-10-19
Fix handling of holes in interpolation for multi-value series

ammojamo:
This PR fixes issue https://github.com/gionkunz/chartist-js/issues/633

I haven't added any new tests, but all existing tests pass. I've added a multi value series to the holes in data examples which demonstrates the fix.

I had to change some logic in core functions `Chartist.isNum` and `Chartist.getNumberOrUndefined` and `Chartist.getMultiValue` because these functions did not handle `null` properly.


gionkunz:
Awesome! This looks perfectly fine. Cheers for the PR and patience!

PR #827, opened 2016-11-01
Force xhtml on span tags

alfmatos:
Span elements under ForeignObject require xhtml namespace to generate valid SVGs.


gionkunz:
Nice! Thanks for the PR!

PR #852, opened 2016-12-09
Fix spelling error

Arinerron:
Error is in word "thikness". Should be "thickness"


gionkunz:
Hi Thanks for the PR! We're generating those HTML files based of the data files located within the website folder. You might do a PR for this file https://github.com/gionkunz/chartist-js/blob/develop/site/data/pages/getting-started.yml

PR #870, opened 2017-01-20
Adding colors support to donut charts

McQuinTrix:
Hi! 

Great Responsive library. One of the reason I started using this library is because it is super responsive. I was until now using Chart.js and was able to add colors to the pie charts too and I thought that was missing in this library. With this commit I am just trying to add colors support, please let me know if you guys feel this is bad-good way to add colors support!

Thanks


gionkunz:
Hi, thanks for the PR! I'm sorry but we really don't want to support this. Chartistians believe in strong separation of concern and colours spread around in your JavaScript is kind of smelly. Check out this page to read more: http://gionkunz.github.io/chartist-js/getting-started.html#colour-up-your-charts

Also for future, please only commit to the source files. Dist files will be generated during releases. Check out the contribution guidelines: https://github.com/gionkunz/chartist-js/blob/develop/CONTRIBUTING.md

Anyway, thanks for the PR, and I hope to see an other one in the future! :-)

PR #871, opened 2017-01-20
Enable drawing donut charts using filled shapes

skovalyov:
* Add donutSolid option support.
* Add solid donut charts examples.

### DIFF
--- a/src/scripts/charts/pie.js
+++ b/src/scripts/charts/pie.js
@@ -26,6 +26,7 @@
       series: 'ct-series',
       slicePie: 'ct-slice-pie',
       sliceDonut: 'ct-slice-donut',
+      sliceDonutSolid: 'ct-slice-donut-solid',
       label: 'ct-label'
     },
     // The start angle of the pie chart in degrees where 0 points north. A higher value offsets the start angle clockwise.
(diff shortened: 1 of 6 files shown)


skovalyov:
By this pull request I would like to revitalize #330, because there was no activity recently. I used the input from the original pull request, implemented maintainer comments and made some changes taking into account the latest code.

src/scripts/charts/pie.js
```diff
@@ -178,21 +183,40 @@
...
-      var path = new Chartist.Svg.Path(!options.donut)
+      var path = new Chartist.Svg.Path(!options.donut || options.donutSolid)
         .move(end.x, end.y)
         .arc(radius, radius, 0, endAngle - startAngle > 180, 0, start.x, start.y);
 
       // If regular pie chart (no donut) we add a line to the center of the circle for completing the pie
       if(!options.donut) {
         path.line(center.x, center.y);
+      } else {
+        if (options.donutSolid) {
```
gionkunz: Can you make an `else if` from this?

gionkunz (commented):
Thanks for the PR! At some point we should also support percentage by using `stripUnit` and calculate relative donut widths. But for the moment this is great! Thanks!

src/scripts/charts/pie.js
skovalyov: Done.

gionkunz:
Awesome, thanks!

Issue #619, opened 2016-02-22
CSP unsafe-inline support


craigfrancis:
Just debugging some JS in the [contrast-widget](https://github.com/AdaRoseEdwards/contrast-widget) for AdaRoseEdwards, and came across:

https://github.com/gionkunz/chartist-js/blob/develop/dist/chartist.js#L339

```
svg = new Chartist.Svg('svg').attr({
  width: width,
  height: height
}).addClass(className).attr({
  style: 'width: ' + width + '; height: ' + height + ';'
});
```

It looks like the `width` and `height` is being set twice, the second one using an inline style attribute (which is considered as [unsafe-inline](https://www.w3.org/TR/CSP2/#directive-style-src) when you're applying a CSP header).

Am I missing something, or can that just be changed to:

```
svg = new Chartist.Svg('svg').attr({
  width: width,
  height: height
}).addClass(className);
```

gionkunz:
Hi there, yeah somehow I missed the comment on this line sorry. The reason for this was a bug in Firefox which did not behave correctly on width and height only set as attributes on the SVG element in some cases. We'd need to re-validate if that's still an issue and remove if it works without the style properties across all supported browsers.

craigfrancis:
Thanks @gionkunz, realise browser support can be difficult... I would be interested to know if/when it was fixed in Firefox, as 44.0.2 seems to be ok without it (well, in this case at least).

gionkunz:
Closing this issue in favor of #620

Issue #620, opened 2016-02-22
CSP review and implement changes


craigfrancis:
Related to Issue #619:

https://github.com/gionkunz/chartist-js/blob/develop/dist/chartist.js#L913

```
if(useForeignObject) {
  // We need to set width and height explicitly to px as span will not expand with width and height being
  // 100% in all browsers
  var content = '<span class="' + classes.join(' ') + '" style="' +
    axis.units.len + ': ' + Math.round(positionalData[axis.units.len]) + 'px; ' +
    axis.counterUnits.len + ': ' + Math.round(positionalData[axis.counterUnits.len]) + 'px">' +
    labels[index] + '</span>';

  labelElement = group.foreignObject(content, Chartist.extend({
    style: 'overflow: visible;'
  }, positionalData));
} else {
  labelElement = group.elem('text', positionalData, classes.join(' ')).text(labels[index]);
}
```

Can we remove the two `style` attributes?

Ideally the width, height, and overflow would be set directly, rather than relying on the HTML/CSS parsing :-)

That said, if you could use `document.createElement('span')`, `.setAttribute('class', ...)`, and `.textContent`, then that really would be magic (tedious to do with an existing code base, but it really helps the security, and sometimes even the performance).

gionkunz:
Same problem here as relying on percentage inside of foreignObjects is an issue for some browsers. Can you open a generic issue for this and #614 to cover both? We can work on that then to identify all issues and find a way to support proper CSP.

gionkunz:
As far as element creating is concerned I believe since Chartist.Svg now also supports rendering of plain HTML elements, we should be able to use it for the element creation so we won't need the content attribute anymore. We still need to investigate though.

craigfrancis:
Hi @gionkunz, not sure what you mean by opening a generic issue? you might be best to do this, as I'm not really familiar with this project (I was just doing some debug, and I really should be getting back to work).

gionkunz:
Well this is an open-source project so everyone is subject to contribute ;-) Thanks for your efforts so far. Let's see where this fits in.

PR #928, opened 2017-04-05
Use HTML DOM style Property instead of adding inline styles to elements

x1c0:
Charts rendering are being blocked by Content Security Policy in the inline style rule.


gionkunz:
Nice! Thanks. This fixes #619 #620

PR #938, opened 2017-04-25
Allow HTML code in labels again (fix #937)

### LINKED ISSUE #937: [Bug] HTML on labels not working anymore since 0.11.0
Since version 0.11.0 the labels cannot have HTML anymore in it.

Reason is this commit:
https://github.com/gionkunz/chartist-js/commit/3857a194b101eb5268469df3c0be6120c40927a3#diff-24396d76ef1bcec041b43d683ecd750bR978

There we now use "innerText". We should change it to "innerHtml" or something else that supports HTML code.

ghost:


gionkunz:
For security reasons we'll keep the default to text. If someone whants to render HTML, they should use a simple plugin: https://github.com/gionkunz/chartist-js/issues/1015

PR #952, opened 2017-05-16
Add 'parameters' attribute on bars

loboulet:
### Why?
I came across an issue where I needed more information on graph bars that `meta` or `value` could provide.

### Solution
Add a new attribute called `parameters` that can be an object or anything the user needs to provide more information on every bar.

I've also renamed `getMetaData`  for `getDataFromSelector` because the path to extract data from `parameters` is the same as the one for `meta`.

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -354,12 +354,14 @@
         positions.y1 = Math.min(Math.max(positions.y1, chartRect.y2), chartRect.y1);
         positions.y2 = Math.min(Math.max(positions.y2, chartRect.y2), chartRect.y1);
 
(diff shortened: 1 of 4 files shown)


gionkunz:
Hey, how comes meta was not sufficient for your use-case? Can you give more details?

loboulet:
Hey,

My use-case is that I have a statistics graph per day. On click of a bar, we need to know which day it is  with some other parameters in order to "drill down" into that day and display informations of that specific day.

I also use the `tooltip` plugin and the `meta` value is used within the `tooltip` so I figured I should add a new possibility to pass information for every bar.

PR #954, opened 2017-05-17
Step Interpolation Improvement

doctormo:
Smaller code, better control with the postpone option to allow half steps. (almost like a bar chart)

### DIFF
--- a/src/scripts/interpolation.js
+++ b/src/scripts/interpolation.js
@@ -389,7 +389,7 @@
    * @returns {Function}
    */
   Chartist.Interpolation.step = function(options) {
-    var defaultOptions = { 
+    var defaultOptions = {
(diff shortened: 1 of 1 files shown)


gionkunz:
Hi there, thanks for your contribution. Can you please describe some use cases for this?

doctormo:
We're using it to render shewhart control chart deviations with a 'square' style. Both upper and lower limits.

What may present a problem for less complex charts, is the points being off the tick marks. Our charts have no points, but we could see that being an issue.

What's the use requirement for code contributions?

doctormo:
I'm attaching a screenshot of the step in use. As you can see, it's used for the upper and lower standard deviation lines (data is not correct, just for illustration).

control-charts-pp

Issue #1022, opened 2017-10-09
Last label not showing


troldmand:
Good morning, beautiful chartists. 

I'm missing the last label on my axisX, and can't figure out why. Data is there, but there's no sign of the last label on the front-end (so it's not just covered by something else, or lacks right-padding). 

Here's the code-snippet I believe is relevant: 

```
        new Chartist.Line('.graph-container.' + patient + '-graph', data, {
            // fullWidth: true,
            chartPadding: {
                top: 0,
                right: 60,
                bottom: 0,
                left: 0
            },
            // lineSmooth: Chartist.Interpolation.none({
            //     fillHoles: true
            // }),
            high: limits.high + 2,
            low: limits.low - 2,
            axisX: {
                type: Chartist.FixedScaleAxis,
                divisor: divisions,
                labelInterpolationFnc: function(value) {
                    console.log("Moment: " + moment(value).format('D/MM'))
                    return moment(value).format('D/MM');
                }
            },
```

On a somewhat related issue, I found [this](http://jsbin.com/nepixe/edit?js,output) snippet, with: 
```
var chart = new Chartist.Line('#chart', {
  labels: data.map((value, index) => `${index + 1}. Data Point`),
  ...
```
... and I wonder, why isn't it showing? And why does the writer in this example go out of his or hers way, to set a 'last label'? 

I should say, that I'm very new to chartist.js.

troldmand:
This chart from the examples page, does the exact same thing. Why isn't the last label showing? 

![screen shot 2017-10-09 at 11 54 33 am](https://user-images.githubusercontent.com/9819978/31333024-dc7c5f82-ace8-11e7-9173-e8da20b1a945.png)

mlumbroso:
Made a PR to fix this behaviour :)

gionkunz:
The momentjs example on the documentation website is simplified a lot. We're still working on a fully functional time scale axis which will arrive with Chartist 1.0. There are two issues with your example. First I believe you're looking for an option to stretch your x-axis to the full width of the chart. However, this option does currently not exist on the FixedScaleAxis. The second problem is the simplification using divisor for time-based axes within the example provided on the doc site. You'd be better off by providing exact tick values for the axis by calculating them from your data using moment date math functions. Here's a working example of how to do that:
```javascript
// Requires Moment.js
var data = {
  series: [
    {
      name: 'series-1',
      data: [
        {x: new Date(143134652600), y: 53},
[...]

PR #1051, opened 2017-12-19
Fix last label not displayed using FixedScaleAxis (#1022)

mlumbroso:
Hi,

First, thanks for making this great lib available.

As reported on #1022 last label was not displayed when we use FixedScaleAxis on x-axis.

Tests are running properly after the change, I think I've followed the guidelines. Let me know if something needs to be changed.

Thanks !


gionkunz:
Hi there. Thanks for your PR! :-) Please check the comment on #1022. I think what would be nice is to add a similar feature like the stretch feature on the AutoScaleAxis. The changes in your PR however, are just adding up to the divisor, which is not the same behaviour as the stretch property of AutoScaleAxis.

PR #1083, opened 2018-04-04
Move away from bower to yarn

Willibaur:
With this change we move away form `bower` to `yarn` as it is recommendend by bower author and explained in his [blog](https://bower.io/blog/2017/how-to-migrate-away-from-bower/).

You can comment on this [issue](https://github.com/gionkunz/chartist-js/issues/1085)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '5.5.0'
+  - '5.7.0'
```
Willibaur, line 3: I had to upgrade node version as `yarn` does not support 5.5.0 and Travis was not happy because of that. 
See [Travis error](https://travis-ci.org/gionkunz/chartist-js/builds/362134261)

image

Willibaur:
There is no intention of merging this change therefore there is no reason to have it opened

PR #1113, opened 2018-07-11
Stacked bar chart: 'accumulate-relative' stackMode

MuffinFox:
I added a stackMode to accumulate negative and positive values separately.

See example for following data series:

series: [
    [5, 4, -3, -5],
    [5, -4, 3, -5]
  ]

chartist_accumulate-relative


alejandrovasta:
+1 on this! I think a new mode is not needed, it should work with the 'accumulate' mode, and doing this if the series contain negative values.

dangreen:
Done in https://github.com/chartist-js/chartist/pull/1375

Issue #876, opened 2017-01-26
Use more robust array checks


CedricReichenbach:
## Problem

In some places objects are checked to be arrays using `obj instanceof Array`. This is problematic since it only compares against the prototype Array in the current context. As a consequence, arrays created in an iframe are not detected as such in top-level code, or vice versa.

See this JSFiddle for an example: http://jsfiddle.net/qchybjp4/1/

## Solution

A more rubust way to check whether an object is an array is `Array.isArray(obj)`, so simply replacing all occurences of `instanceof Array` should suffice.

## Background

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/instanceof

StK88:
+1

PR #1128, opened 2018-09-11
Replace `instanceof Array` for `Array.isArray()`.

sinspiral:
This should fix the #876 I was running into with iframes and arrays context.


hansmaad (approved):

dangreen:
Done in https://github.com/chartist-js/chartist/commit/ee4106e04f3c081805dd79675340378f895c8290

Issue #1095, opened 2018-05-09
Allow overriding responsiveOptions with update


mestaritonttu:
There is an issue in react-chartist for this, offering some workarounds: https://github.com/fraserxu/react-chartist/issues/38

Any other workarounds we can use before this is (hopefully) implemented?

Edit: I tried hacking it in chartist.js like this, but it is not working:

`function update(data, options, responsiveOptions, override) {`

...

```
if(responsiveOptions) {
  this.responsiveOptions = Chartist.extend({}, override ? this.responsiveOptions : this.responsiveOptions, responsiveOptions);
}
```

PR #1129, opened 2018-09-12
Add optional param to `update` to customize the `responsiveOptions`

sinspiral:
This should allow customizing the responsive options as requested on the #1095


src/scripts/base.js
```diff
@@ -35,6 +36,10 @@
       });
     }
 
+    if(responsiveOptions) {
+      this.responsiveOptions = responsiveOptions;
```
matteoraf, line 40: Shouldn't this be like:
`this.responsiveOptions = Chartist.extend({}, override ? this.responsiveOptions : [], responsiveOptions);`

PR #1143, opened 2018-11-20
Use HTML DOM style for setting label overflow as well

adamchainz:
PR #928 fixed most inline style errors from Content Security Policy, but it missed this one. This fixes another error by again using `style` to add the style rule.


adamchainz:
This is the error I was seeing on my application:
screen shot 2018-11-20 at 18 05 56

Building with this patch fixes it.

PR #1195, opened 2019-10-28
Bugfix: only calculate high low from stacked values if stackMode is accumulate

dannyvankooten:
When not accumulating totals in stacked bar charts, the high / low values should not do so either.

Turns this:
afbeelding

Into this:
afbeelding


gionkunz:
Hey thanks for that, this looks absolutely legit! Let me take a better look
this weekend. Thanks!

Cheers
Gion

Danny van Kooten <notifications@github.com> schrieb am Mo., 28. Okt. 2019,
15:49:

dannyvankooten:
That'd be awesome @gionkunz. Thank _you_ for the great library!

PR #1237, opened 2020-04-11
docs: Fix simple typo, unncessary -> unnecessary

timgates42:
There is a small typo in dist/chartist.js, src/scripts/charts/pie.js.

Should read `unnecessary` rather than `unncessary`.


dangreen:
Will resolved in https://github.com/chartist-js/chartist/pull/1339

PR #1255, opened 2020-06-18
Bugfix: summing up stacked bars with undefined items in series

### LINKED ISSUE #1254: Bug in summing up stacked bars with undefined items in series
When there are `undefined` items in the series, the summing of stacked bars goes wrong.
I tracked it to this line:
https://github.com/gionkunz/chartist-js/blob/e7e78201bffe9609915e5e53cfafa29a5d6c49f9/src/scripts/charts/bar.js#L152

Instead of 
```
return {
    x: prev.x + (curr && curr.x) || 0,
    y: prev.y + (curr && curr.y) || 0
};
```
It should be
```
return {
    x: prev.x + ((curr && curr.x) || 0),
    y: prev.y + ((curr && curr.y) || 0)
};
```
Otherwise whenever `curr` is undefined, the `reduce` resets to 0.

JanMisker:
When there are undefined items in the series, the summing of stacked bars goes wrong. 

Fixes #1254

### DIFF
--- a/src/scripts/charts/bar.js
+++ b/src/scripts/charts/bar.js
@@ -148,8 +148,8 @@
           return value;
         }).reduce(function(prev, curr) {
           return {
-            x: prev.x + (curr && curr.x) || 0,
-            y: prev.y + (curr && curr.y) || 0
+            x: prev.x + ((curr && curr.x) || 0),
+            y: prev.y + ((curr && curr.y) || 0)
           };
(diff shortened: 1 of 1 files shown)


gionkunz:

Arantiryo:
I think this PR can be closed: https://github.com/chartist-js/chartist/issues/1254#issuecomment-1250114191

PR #1260, opened 2020-07-29
create new Stack chart

Zadvornyi:


gionkunz:

Zadvornyi:
Thanks for your answer! I will rewrite my commit

PR #1264, opened 2020-08-09
Add types from DefinitelyTyped locally.

tjni:
Because we have forked chartist and intend to update it, we also need
a copy of its types to allow us to change them.


tjni:
Whoops, I totally messed this up when setting up my fork. If I have useful changes, I'll create PRs for them upstream.

PR #1323, opened 2022-08-06
chore: changed publishing approach

gionkunz:
I read through https://github.com/webpack/webpack/issues/4674 and changed the publishing approach accordingly.


dangreen:
@gionkunz Hi. Webpack should receive ESM treeshakable bundle , but now it will use iife bundle from browser field. Also it's not UMD bundle but named like that. UMD bundle builds for main field (you can see it in Rollup config). I sugges you to revert this PR.

PR #1327, opened 2022-08-07
docs: fix img url

dangreen:


gionkunz (approved):
We will then also change it again and probably include it in the website project ir the main branch once we delete master.

PR #1352, opened 2022-08-29
docs: new website

dangreen:
To run locally:

```bash
cd website
pnpm i
pnpm start
```


gionkunz:
Hi @dangreen thanks for the PR. I was thinking more of Gatsby SSG. Maybe we can still discuss this? Also, do you think we should have this in the same repo? If yes, we should probably use workspaces or even lerna so we have a bit more monorepo tooling at hand. I personally use Nx in my projects but then this would replace the whole build stack (Nx is based on Webpack and only uses Rollup for specific outputs I think). Or we go poly repo and create chartist/website

dangreen:
@gionkunz Hi. Docusaurus is also SSG. On the build step, it reads sources to generate API docs using [typedoc](https://github.com/TypeStrong/typedoc), so it should be in the same repo. If we will use worksapces, then `pnpm install` in the root will also install deps in `website/`, what not always required. CI steps will be faster without it. CI steps that work with the website will install deps in `website/` separately.

igorlukanin:
@gionkunz I would probably stick with what @dangreen suggests if you don't have a really strong opinion about this.

gionkunz:
What about search? Can you easily generate a search index and is there an easy way to integrate a local search like with Gatsby?

dangreen:
@gionkunz Yes, with [Algolia](https://docusaurus.io/docs/search#using-algolia-docsearch).

gionkunz:
@dangreen just tried to test it and I get a lot of errors. Somehow the spec files get read and the processes freaks out from the global expect:

```
Error: ../src/core/data/bound.spec.ts:98:7 - error TS2582: Cannot find name 'it'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.

98       it('should return single step if range is less than smallest increment', () => {
         ~~

Error: ../src/core/data/bound.spec.ts:105:9 - error TS2304: Cannot find name 'expect'.

105         expect(bounds.min).toBe(999.9999999999999);
            ~~~~~~

Error: ../src/core/data/bound.spec.ts:106:9 - error TS2304: Cannot find name 'expect'.

106         expect(bounds.max).toBe(1000);
            ~~~~~~

Error: ../src/core/data/bound.spec.ts:107:9 - error TS2304: Cannot find name 'expect'.

107         expect(bounds.low).toBe(999.9999999999997);
            ~~~~~~

Error: ../src/core/data/bound.spec.ts:108:9 - error TS2304: Cannot find name 'expect'.

108         expect(bounds.high).toBe(1000.0000000000001);
            ~~~~~~

Error: ../src/core/data/bound.spec.ts:109:9 - error TS2304: Cannot find name 'expect'.

109         expect(bounds.values).toEqual([roundWithPrecision(999.9999999999999)]);
            ~~~~~~

Error: ../src/core/data/normalize.spec.ts:3:1 - error TS2582: Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.
```

dangreen:
@gionkunz Hi. Did you install dependencies in the root and in the website dirs?

gionkunz:
Also, I saw that you're referencing CUBE Slack as a chat for Chartist. I think we should use Github Discussions and the Chartist Gitter Chat instead: https://gitter.im/gionkunz/chartist-js

dangreen:
@gionkunz fixed

gionkunz:
> @gionkunz Hi. Did you install dependencies in the root and in the website dirs?

I did what you wrote just pnpm i in the website folder. Do I need to do it on the root too?

dangreen:
@gionkunz yes

gionkunz:
Hi @dangreen I was able to install the website now. Still, it feels a bit weird that the website is just a subfolder of the project. I think it would be much better if we create a pnpm workspace and have the library separated from the website. It should be possible to reference the files from a different workspace package to parse the TS files to create the API docs right? In nx that's also possible, and there's even a docusaurus community plugin. There we have apps/ui, apps/docs etc.

I also think that we will need some theming to bring that chartist 1 spirit over to chartist 2. It will lose a bit of heart and soul if we just use the standard docusaurus theme. Of course this can be done in a second step.

Other issues I have found:
- The code references (view source) still refer to the old repo
- The codesandboxes don't work (maybe also an issue with codesandbox, as I cannot access the website at the moment)
- On the quickstart we're still referring to the CUBE Slack chat
- For the domain of the new website we can use chartist.dev

dangreen:
@gionkunz 

> create a pnpm workspace

Done

> need some theming 

let's make it as separate task

> The code references (view source) still refer to the old repo

I'm not sure what you mean

> The codesandboxes don't work

I think it is a problem on your side, cause all works fine for me.

> On the quickstart we're still referring to the CUBE Slack chat

Fixed

> For the domain of the new website we can use chartist.dev

Are you own this domain?
js.org domains are free for js open source projects.

gionkunz:
Thanks for the changes!

Yes, I already own the domain chartist.dev, so we can use it.

About the "view source" links, this is what I mean:
image

dangreen:
@gionkunz I think it depends on your local git repo config. In CI env all will be ok.

Снимок экрана 2022-09-11 в 13 19 13

igorlukanin:
@gionkunz I hope this can be merged soon. Can't wait to see the new docs website live! 👀

igorlukanin (approved):
LGTM!

gionkunz:
I think technically it looks good. However, I dont think we should go live with it until we make it look a bit more like the charist "brand".

gionkunz:
We can merge it but i would only publish it once we have it look like a new version of the old documentation website. Also the getting started guide needs to be migrated and other thibgs which are currently there on the old website. We can add this later but I think its very important.

Issue #1248, opened 2020-05-25
Svg.Path.parse fails on negative values


hansmaad:
Hey Gion,

the `Svg.Path.parse` function fails if the path contains negative values like
```d="M-10,10L-100,-100"```

This line
https://github.com/gionkunz/chartist-js/blob/e7e78201bffe9609915e5e53cfafa29a5d6c49f9/src/scripts/svg-path.js#L192

must probably be changed to
```var chunks = path.replace(/([A-Za-z])(-?[0-9])/g, '$1 $2') ```

gionkunz:
Hi Hannes. Hope youre doing well. Thanks for the bug report. I have not touched Chartist for quite some time since I'm very busy with other things. Would appreciate a PR if you or anyone would have time.

PR #1360, opened 2022-09-25
fix: SvgPath.parse fails on negative values

Arantiryo:
Implements suggested fix from #1248.


gionkunz:
Hi there! Thanks for the PR! Would you mind adding a few tests for this case? Cheers!

Arantiryo:
Hi @gionkunz! Yes, absolutely! Do you think that covers it or should I add anything else?

gionkunz (approved):

gionkunz:
Thanks, @Arantiryo  , I think that's sufficient at the moment. Cheers!

PR #1392, opened 2022-12-26
feat: add an error message when chart container is not found

### LINKED ISSUE #1354: [Bug]: More descriptive error message when DOM node for container was not found
### Would you like to work on a fix?

- [ ] Check this if you would like to implement a PR, we are more than happy to help you go through the process.

### Current and expected behavior

Currently, when Chartist is used in dynamic rendering situations or the wrong selector is passed to initialize Chartist, we are having a very unspecific error message directly from the DOM API:

```
[ERROR TypeError: Cannot read property 'querySelectorAll' of null]
```

We should catch this error and enrich it with better description of what went wrong so users know how to fix it.

### Reproduction

Initialize Chartist with an invalid selector

### Chartist version

v2.x

### Possible solution

Enrich the error message

Arantiryo:
Fixes #1354.

Error message example:
no-container-error-chartist

### DIFF
--- a/src/charts/BaseChart.ts
+++ b/src/charts/BaseChart.ts
@@ -32,7 +32,9 @@ export abstract class BaseChart<TEventsTypes = BaseChartEventsTypes> {
 
     if (!container) {
       throw new Error(
-        `Target element ${typeof query === 'string' ? `"${query}"` : ''} is not found`
+        `Target element ${
+          typeof query === 'string' ? `"${query}"` : ''
+        } is not found`
       );
     }
 
--- a/src/charts/BaseChart.ts
+++ b/src/charts/BaseChart.ts
@@ -32,7 +32,7 @@ export abstract class BaseChart<TEventsTypes = BaseChartEventsTypes> {
 
     if (!container) {
       throw new Error(
-        `Target element ${typeof query === 'string' ? query : ''} is not found`
+        `Target element ${typeof query === 'string' ? `"${query}"` : ''} is not found`
       );
     }
 
--- a/src/charts/BaseChart.ts
+++ b/src/charts/BaseChart.ts
@@ -31,7 +31,9 @@ export abstract class BaseChart<TEventsTypes = BaseChartEventsTypes> {
       typeof query === 'string' ? document.querySelector(query) : query;
 
     if (!container) {
-      throw new Error('Target element is not found');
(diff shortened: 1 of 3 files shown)


gionkunz:
Awesome thats long overdue 😂👍

gionkunz:
In strict typescript container always needs to be defined. I would anyway catch the error earlier before this function call because then you can also print hinting information like the selector that was used.

Arantiryo:
Good suggestion @gionkunz, thanks for pointing that out! 

Since we can't catch get the selector right before this function call I figured we could modify the error message in the `BaseChart.ts` so the both error messages would look like this:

no-container-error-msg

What do you think?

gionkunz:
Cool! We had many issues open because people did not understand this error.

What about the size limit check? Do we need to adjust the threshold?

Arantiryo:
Probably. Let me kindly ask @dangreen about this, he knows way more about CI / CD than I do. :)

src/charts/BaseChart.ts
```diff
@@ -31,7 +31,9 @@ export abstract class BaseChart<TEventsTypes = BaseChartEventsTypes> {
       typeof query === 'string' ? document.querySelector(query) : query;
 
     if (!container) {
-      throw new Error('Target element is not found');
+      throw new Error(
+        `Target element ${typeof query === 'string' ? query : ''} is not found`
+      );
```
dangreen:
```suggestion
      throw new Error(
        `Target element ${typeof query === 'string' ? `"${query}"` : ''} is not found`
      );
```

src/core/creation.ts
```diff
@@ -27,6 +27,9 @@ export function createSvg(
   height: number | string = '100%',
   className?: string
 ) {
+  if (!container) {
+    throw new Error('Container element is not found');
+  }
   // Check if there is a previous SVG element in the container that contains the Chartist XML namespace and remove it
```
dangreen:
```suggestion
  if (!container) {
    throw new Error('Container element is not found');
  }

  // Check if there is a previous SVG element in the container that contains the Chartist XML namespace and remove it
```

dangreen (dismissed):
@Arantiryo need to edit `.size-limit.json` file, run `pnpm test:size` to run size-limit

gionkunz (approved):

PR #1403, opened 2023-03-02
feat(SVG): support for viewport in SVG only providing width and height

### LINKED ISSUE #1402: Support for viewBox for scalable and responsive charts
### Would you like to work on this feature?

- [X] Check this if you would like to implement a PR, we are more than happy to help you go through the process.

### What problem are you trying to solve?

I am aware that the advantage of charist is its responsiveness yet I would be more than glad to control the viewport allowing to create resizeable charts until a certain breakpoint. Because I created some Plugins for Legend, Hover, Highlights and Titles they tend to conflict in resizing.

### Describe the solution you'd like

This can be an optional value in Options that could also be controled with resonsive options later on adjusting the viewport for mobile screens.

I would be more than happy to create a PR.

### Describe alternatives you've considered

I tried different approaches using CSS classes with Media-Queries but they never created a reliable result when resizing in mobile resolutions.

### Documentation, Adoption, Migration Strategy

_No response_

evangelion1204:
For now a draft but this should outline what should be achieved. The charts scale without redrawing and rearranging. The drawback might be that fonts can get too small if no responsive options are used.

I do believe we could also smartly skip redraw in such cases until we hit a breakpoint.

Closes #1402

### DIFF
--- a/src/charts/BarChart/BarChart.stories.ts
+++ b/src/charts/BarChart/BarChart.stories.ts
@@ -445,3 +445,46 @@ export function AccumulateRelativeStack() {
 
   return root;
 }
+
+export function ViewBox() {
+  const root = document.createElement('div');
+
+  new BarChart(
+    root,
+    {
+      series: [
+        [
+          { x: 1, y: 1 },
+          { x: 3, y: 5 }
+        ]
+      ]
+    },
+    {
+      axisX: {
+        type: AutoScaleAxis,
+        onlyInteger: true
+      },
+      axisY: {
+        type: AutoScaleAxis,
+        onlyInteger: true
+      },
+      viewBox: {
+        width: 800,
+        height: 300
+      }
+    },
+    [
+      [
+        'screen and (max-width: 575px)',
+        {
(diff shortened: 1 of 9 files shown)


gionkunz:
Hey, thanks for your contribution! I guess this would also be relevant for the other chart types, right? https://github.com/chartist-js/chartist/blob/main/src/charts/BarChart/BarChart.ts#L215-L221

evangelion1204:
Sure, I just use a fork right now and needed it for Line-Charts although I also have Bar-Chats in use. I am going to add it here too. I will update the PR within the next days.

evangelion1204:
@gionkunz the example I added it to is not too great though. Only thing missing are unit tests around the `createSvg` function.

evangelion1204:
Ok, updated the tests.

evangelion1204:
@gionkunz In case you got some time please review. I would also like to create more PRs later on.

evangelion1204:
@gionkunz Quick reminder 😄

dangreen (approved):

PR #1433, opened 2024-10-02
fix: prototype pollution vulnerability in extend (CVE-2024-45435)

### LINKED ISSUE #1427: [Bug]:  Prototype Pollution Vulnerability Affecting chartist  module, versions >=1.0.0 <=1.3.0
### Would you like to work on a fix?

- [ ] Check this if you would like to implement a PR, we are more than happy to help you go through the process.

### Overview

A Prototype Pollution vulnerability Affecting chartist , versions >=1.0.0 <=1.3.0, due to missing check if the argument resolves to the object prototype. This allow the attacker to inject malicious object property using the built-in Object property `__proto__` which recursively assigned to all the objects in the program.

### Reproduction

_sent directly to the maintainer's email_

### Chartist version

1.3.0

### Possible solution

> + Freeze the root prototype using Object.freeze
> + Require schema validation of JSON input.
> + Avoid using unsafe recursive merge functions.
> + Consider using objects without prototypes (for example, Object.create(null)), breaking the prototype chain and preventing pollution.
> + As a best practice use Map instead of Object

andersk:
- Fixes #1427.

https://nvd.nist.gov/vuln/detail/CVE-2024-45435
https://gist.github.com/tariqhawis/c67177164d3b7975210caddb25b60d62


tariqhawis:
Hi @andersk ,

Thank you for your feedback regarding the report. I would like to highlight an additional prototype accessor: `constructor.prototype`. A potential injection could look like this:

`extend ({}, JSON.parse('{"constructor":{"prototype":{"polluted":yes}}}'))"`

Checking for the presence of `constructor` in the input should be sufficient. 

Best,
Tariq

andersk:
> Checking for the presence of `constructor` in the input should be sufficient.

Nope, it’s not.

```js
extend({}, {"hasOwnProperty": {"polluted": "yes"}});
console.log(Object.prototype.hasOwnProperty.polluted); // → yes
```

I’ve pushed a more complete fix.

praveen-zensar (approved):

praveen-zensar:
**Please merge this fix as it's a severity score is high.**

src/utils/extend.ts
```diff
@@ -11,7 +11,11 @@ export function extend<T, A, B>(target: T, a: A, b: B): T & A & B;
 export function extend(target: any = {}, ...sources: any[]) {
   for (let i = 0; i < sources.length; i++) {
     const source = sources[i];
+    const targetProto = Object.getPrototypeOf(target);
     for (const prop in source) {
+      if (targetProto !== null && prop in targetProto) {
```
danielpalme, line 16: I propose the following change for this line:  
`if (prop === "__proto__" || prop === "constructor" || (targetProto !== null && prop in targetProto)) {`

See: https://codeql.github.com/codeql-query-help/javascript/js-prototype-pollution-utility/

dangreen (approved):

src/utils/extend.ts
```diff
@@ -11,7 +11,11 @@ export function extend<T, A, B>(target: T, a: A, b: B): T & A & B;
 export function extend(target: any = {}, ...sources: any[]) {
```
dangreen, line 11: All usage cases are where first argument is empty object. We can rename `extend` to `merge`, remove first argument, and as target use `Object.create(null)`.

I can fix it by myself later on next week

dangreen (changes requested):

src/utils/extend.ts
mihaileu: seems that @andersk  might not be able to help here

andersk: No I agree in principle, but here’s the problem: `Chartist.extend` is part of the public API of this package, and if we bump the semver major version to remove it, that’s going to take years to propagate through the ecosystem…

src/utils/extend.ts
andersk: Your change isn’t necessary because both `__proto__` and `constructor` are already skipped by `targetProto !== null && prop in targetProto`.

dangreen:
@andersk You should run "update storyshots" action manually in your fork. Then download artifacts and update screenshots in your branch.

andersk:
@dangreen I did that and it resulted in no changes.

https://github.com/andersk/chartist/actions/runs/14316237387/job/40122907313
https://github.com/andersk/chartist/actions/runs/14316237387/artifacts/2897050220

dangreen:
@andersk 
put all images from https://github.com/andersk/chartist/actions/runs/14316237387/artifacts/2897050220
to https://github.com/andersk/chartist/tree/main/test/__image_snapshots__

andersk:
Never mind, figured it out, fixed.

dangreen (approved):

dangreen:
https://github.com/chartist-js/chartist/releases/tag/v1.3.1

PR #1453, opened 2025-10-28
Add a RadarChart 

leosutic:
This PR adds a `RadarChart`, which is a polar line chart. In order to support polar charts, the guts of the `Axis` class was factored out into a `CartesianAxis` class from which `AutoScaleAxis`, etc. inherit, and a parallel hierarchy was created for `PolarAxis`. The `LineChart` was then cloned and reworked into the `RadarChart`.


gionkunz:
Thank you so much for your contribution! LGTM. I will need some time for a detailed review and testing.

leosutic:
Hi @gionkunz , you're welcome, and thank you for starting this project!

I don't know if the PR is ready for review, but now that I have you here, these are the points that I I have left:

* Can't get Storybook to run due to dependency conflicts
* Can't get the sandboxes to run, would like to use that to test how various parameters affect the chart
* I'm still a bit confused regarding the chart layout, and I'm not sure I'm adding up all the offsets and paddings right.
* Need to test various types of "line" chart (area, point, etc)
* Still not sure how to deal with, for example, `GridDrawEvent`:s for polar grid lines that are circles - the `x1, y1, x2, y2` positional data doesn't really work for circles, so that might require some work.

